PR #3, opened 2021-09-27
Add CI

FabijanC:
- Modeled after sourcify's config.yml with some optimizations:
  - parsing package.json with `jq` (circleci images already have it installed)
  - the publishing job will first check if it's necessary to publish, then set npm auth token, install and build
- Currently there are no tests:
  - they will soon be added (in a different PR)
  - `npm run test` simply calls an `echo` which informs about this
- Pending: `NPM_TOKEN` variable which could be added to [the Shard Labs context](https://app.circleci.com/settings/organization/github/Shard-Labs/contexts/f7f363c6-f101-4ac8-9193-343c88da5fb0).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,48 @@
+version: 2.1
+
+aliases:
+  - &node-base
+    steps:
+      - checkout
+      - run:
+        name: Install
+        command: npm install
+      - run:
+        name: Build
+        command: npm run build
+      - run:
+        name: Test
+        command: npm run test
+
+jobs:
+  test_node_v12:
+    <<: *node-base
+    docker:
+      - image: circleci/node:12
+  test_node_v14:
+    <<: *node-base
+    docker:
+      - image: circleci/node:14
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+      - image: circleci/node:14
+    steps:
+      - checkout
+      - run:
+        name: Publish
+        command: ./scripts/npm-publish.sh
+
+workflows:
+  version: 2.1
+  workflow1:
```
Jakic007, line 37: Can you rename this workflow to "test_and_publish" please?

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+    jobs:
+      - test_node12
```
Jakic007, line 39: Rename to "test_node_v12" please.

Jakic007 (changes requested):
Once you rename these jobs, feel free to merge.

PR #8, opened 2021-10-05
Upgrade to v0.1.6

### LINKED ISSUE #7: [feature] Support `getContractAt`
Hi, is it possible that the plugin supports accessing a deployed Cairo contract similar to how `ethers.getContractAt` works?

Also, I would get `Property 'address' is private and only accessible within class 'StarknetContract'` error if write the code in typescript. Wonder if that can be fixed.

FabijanC:
- Projects using this plugin couldn't use all features if their projects are using typescript; now it's fixed:
  - argument array for starknet function invocation is now `any`
  - contract `address` property is public
- Use a contract factory approach:
  - Use `getContractFactory` instead of `getStarknetContract` (example in README.md)
  - Allow the choice of deploying a new contract or reusing an old address
  - Fix #7


Jakic007 (changes requested):
Can you please remove this commit: f068a71ca2fa438c73ab7b3b9b74d54484236ff0 
Otherwise, looks good!

PR #19, opened 2021-11-12
Upgrade to v0.3.0

FabijanC:
## Usage
- Enable passing constructor arguments (possible to pass them when calling `deploy` of a contract factory).
- Enable signing transactions.
- Rely on JS `BigInt` because Cairo's `felt` type might be too big for ordinary JS `Number`.
- Enforce typed/structured input and output (no more single-string function results; no more blindly passing sequences of numbers as function inputs).
- Fix Mac (OS X) support (when interacting with gateways listening on localhost, such as the devnet).
- Bump version to v0.3.0.

## Examples
These examples provide an overview of breaking changes that affect testing. Find more examples in README.md.

### Example 1 - Array
Assume a contract with function:
```text
func sum_array(a_len: felt, a: felt*) -> (res: felt);
```
What used to look like:
```typescript
const a = [1, 2, 3];
const resStr = await contract.call("sum_array", [a.length, ...a]); // separate length handling was required
const res = parseInt(resStr); // dangerous in case of losing precision with JS Number
chai.expect(res).to.equal(6);
```
now looks like:
```typescript
const { res } = await contract.call("sum_array", { a: [1, 2, 3] }); // no length handling required
chai.expect(res).to.equal(BigInt("6")); // or chai.expect(res).to.equal(6n)
```

### Example 2 - Tuple
Assume a contract with function:
```text
func double_pair(pair: (felt, felt)) -> (res_pair: (felt, felt))
```
What used to look like:
```typescript
const pair = [10, 20];
const resStr = await contract.call("sum_pair", [...pair]);
chai.expect(resStr).to.equal("20 40");
```
now looks like:
```typescript
const { res } = await contrat.call("sum_pair", { pair: [10, 20] });
chai.expect(res).to.equal([20n, 40n]); // 20n is syntactic sugar for BigInt("20")
```

### Example 3 - Struct
Assume a contract with struct and function:
```text
struct Point:
    member x: felt
    member y: felt
end

func sum_points(p1: Point, p2: Point) -> (res: Point)
```
What used to look like:
```typescript
const resStr = await contract.call("sum_points", [1, 2, 3, 4]); // Point(1, 2) + Point(3, 4)
chai.expect(resStr).to.equal("4 6"); // Point(4, 6)
```
now looks like:
```typescript
const { res } = await contract.call("sum_points", { p1: { x: 1, y: 2 } , p2: { x: 3, y: 4 } });
chai.expect(res).to.deep.equal({ x: 4n, y: 6n });
```

## Dev
- Update dependencies.
- Tests in CI are set to use branch `plugin` in `starknet-hardhat-example`. Previously they were using `master`, so it was sometimes hard to synchronize the repository's demonstrative and testing purpose. Once this PR is merged, a merge of `plugin` into `master` in that repository will be required (and subsequent rebases of other branches).

## Review
- @octate I'm assigning you to check if interaction with devnet is supported on your Mac.


Tiago-shardlabs (approved):

octate (approved):
Yes, deployment works on mac @FabijanC

PR #25, opened 2021-12-03
Voyager contract verification

dribeiro-ShardLabs:
- Implemented "starknet-verify" task, which queries Voyager for contract verifications
- Implemented "wait" flag for the "starknet-deploy" task, which if present waits until the status in the deployment transaction for a contract is "Pending"


test/starknet-verify/hardhat.config.ts
```diff
@@ -0,0 +1,7 @@
+import "../dist/index.js";
+
+module.exports = {
+    paths: {
+        starknetArtifacts: "my-starknet-artifacts"
```
FabijanC, line 5: why not just use the default starknetArtifacts path? like in the `plain` test case

test/with-wait/hardhat.config.ts
```diff
@@ -0,0 +1,7 @@
+import "../dist/index.js";
+
+module.exports = {
+    paths: {
+        starknetSources: "my-starknet-sources"
```
FabijanC, line 5: why not just use the default starknetSources?

src/index.ts
```diff
@@ -402,3 +451,75 @@ extendEnvironment(hre => {
...
+            if(args.starknetNetwork === ALPHA_MAINNET)
+                voyagerUrl = VOYAGER_MAINNET_CONTRACT_API_URL;
+            else{
+                const msg = `Unknown starknet-network provided: ${args.starknetNetwork}`;
+                throw new HardhatPluginError(PLUGIN_NAME, msg);
+            }
+        }
+        voyagerUrl+=args.address + "/code";
+        var isVerified = false;
+        await axios.get(voyagerUrl,{
```
FabijanC, line 472: Why use both await and callbacks?
How about this:
```typescript
const resp = await axios.get(url, args);
const data = resp.data;
```
Or something like this: https://github.com/axios/axios#typescript

src/index.ts
```diff
@@ -402,3 +451,75 @@ extendEnvironment(hre => {
...
+                isVerified = true;
+            }
+        })
+        .catch(error => {
+            const msg = `Something went wrong when trying to verify the code at address ${args.address}`;
+            console.log(error);
+            throw new HardhatPluginError(PLUGIN_NAME, msg);
+        });
+        
+        if(isVerified)
```
FabijanC, line 490: Should the user be notified about contract already having been verified?
Also indent the return statement and put it in curly braces.

src/index.ts
```diff
@@ -342,11 +345,57 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
...
+                        tx_hashes.push(executed.stdout.toString().match(".*Transaction hash: (?<tx_hash>\\w*).*").groups["tx_hash"]);
+                    
+                    return execResult;
+                } 
+                else
+                    return processExecuted(executed);
             });
         }
 
+        if(args.wait){      //  If the "wait" flag was passed as an argument, check the previously stored transaction hashes for their statuses
```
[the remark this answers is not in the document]
FabijanC: Ok, probably not until we do the refactor which will remove the recursion.

src/index.ts
```diff
@@ -402,3 +451,75 @@ extendEnvironment(hre => {
...
+        return;
+        //If contract hasn't been verified yet, do it
+        var contract_path = args.path;
+        if (!path.isAbsolute(contract_path)) {
+            contract_path = path.normalize(path.join(hre.config.paths.root, contract_path));
+        }
+        if (fs.existsSync(contract_path)) {
+            const content = {code:fs.readFileSync(contract_path).toString().split(/\r?\n|\r/)};
+            await axios.post(voyagerUrl,JSON.stringify(content))
+            .then(response => {
```
FabijanC, line 500: In case you stick with callbacks, this `response` is unused.

src/index.ts
```diff
@@ -402,3 +451,75 @@ extendEnvironment(hre => {
...
 });
+
+task("starknet-verify", "Verifies the contract in the Starknet network.")
+    .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
+    .addParam("path", `The path of the cairo contract (e.g. contracts/conract.cairo)`)
+    .addParam("address", `The address where the contract is deployed`)
+    .setAction(async (args, hre) => {
+        var voyagerUrl = VOYAGER_GOERLI_CONTRACT_API_URL;
+        
+        if(args.starknetNetwork && args.starknetNetwork !== ALPHA){
```
FabijanC, line 462: Could you make this if-else branching clearer?

src/index.ts
```diff
@@ -342,11 +345,57 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
...
             });
         }
 
+        if(args.wait){      //  If the "wait" flag was passed as an argument, check the previously stored transaction hashes for their statuses
+            let done = false;
+            var hash_done:boolean [] = [];
+            for(var i = 0; i<tx_hashes.length;i++){
+                hash_done.push(false);
+            }
+            while(!done){   // Run loop until all transactions are PENDING or ACCEPTED
```
FabijanC, line 367: This action of the starknet-verify task is quiet big. Some of its sections should be extracted to separate functions.

package-lock.json
```diff
@@ -1,4062 +1,8 @@
 {
```
FabijanC, line 1: Perhaps this is a matter of npm version you're using, but it'd be the best if we both use a version that generates a package-lock.json with lockfileVersion 2.

FabijanC (changes requested):
The new changes seem to cover my comments from the last review, except the required changes to package-lock.json: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/25#discussion_r762091241

FabijanC (approved):

Tiago-shardlabs (approved):

src/types.ts
```diff
@@ -5,6 +5,7 @@ import { PLUGIN_NAME, CHECK_STATUS_TIMEOUT } from "./constants";
 import { adaptLog, adaptUrl } from "./utils";
 import { adaptInput, adaptOutput } from "./adapt";
 import { StarknetWrapper } from "./starknet-wrappers";
+import { Console } from "console";
```
FabijanC, line 8: This line is probably accidentally here.

src/index.ts
```diff
@@ -349,7 +349,7 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
                     const execResult = processExecuted(executed);
                     
                     if(execResult == 0)
-                        tx_hashes.push(executed.stdout.toString().match(".*Transaction hash: (?<tx_hash>\\w*).*").groups["tx_hash"]);
+                    txHashes.push(executed.stdout.toString().match(".*Transaction hash: (?<tx_hash>\\w*).*").groups["tx_hash"]);
```
FabijanC, line 352: I guess you didn't want this line to look the way it does. Perhaps you meant to indent it and enclose with curly braces. Also, would it be possible to utilize `extractTxHash` from types.ts?

src/index.ts
```diff
@@ -5,11 +5,11 @@ import { task, extendEnvironment, extendConfig } from "hardhat/config";
 import { HardhatPluginError } from "hardhat/plugins";
 import { ProcessResult } from "@nomiclabs/hardhat-docker";
 import "./type-extensions";
-import { StarknetContractFactory, checkStatus, isTxAccepted } from "./types";
+import { StarknetContractFactory, iterativelyCheckStatus } from "./types";
 import { PLUGIN_NAME, ABI_SUFFIX, DEFAULT_STARKNET_SOURCES_PATH, DEFAULT_STARKNET_ARTIFACTS_PATH, DEFAULT_DOCKER_IMAGE_TAG, DOCKER_REPOSITORY, DEFAULT_STARKNET_NETWORK, ALPHA_URL, ALPHA_MAINNET_URL, VOYAGER_GOERLI_CONTRACT_API_URL, VOYAGER_MAINNET_CONTRACT_API_URL, ALPHA_MAINNET, ALPHA} from "./constants";
 import { HardhatConfig, HardhatRuntimeEnvironment, HardhatUserConfig, HttpNetworkConfig } from "hardhat/types";
 import { adaptLog, adaptUrl, getDefaultHttpNetworkConfig } from "./utils";
-import { DockerWrapper, VenvWrapper } from "./starknet-wrappers";
+import { DockerWrapper, StarknetWrapper, VenvWrapper } from "./starknet-wrappers";
```
FabijanC, line 12: `StarknetWrapper` is not explicitly used.

FabijanC (changes requested):

FabijanC (approved):

PR #26, opened 2021-12-07
Change contract path handling

dribeiro-ShardLabs:
Removed the recursive function "traverseFiles" and changed it for "glob"
Added a check for faulty venv 
Added support for relative paths for contractFactory


src/index.ts
```diff
@@ -386,14 +374,20 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
...
+        paths = await glob(path.join(traversable,"**","*"));
+    }
+    else{
+        paths.push(traversable);
+    }
+    paths.filter(file => {
+        return path.basename(file) === name;
+    }).forEach(file => {
+        foundPath = file;
+        return 0;
```
FabijanC, line 388: No need to return the `0`. That used to be so just to comply with what `traverseFiles` expected - it expected a number as a return value, so it was given a symbolic zero.
Now that I think about it, this is how it should be implemented: it should be checked whether `paths` contains just one path, throw if it is not so (with a message saying it's ambiguous), and return the zeroth element.

src/index.ts
```diff
@@ -10,31 +10,8 @@ import { PLUGIN_NAME, ABI_SUFFIX, DEFAULT_STARKNET_SOURCES_PATH, DEFAULT_STARKNE
 import { HardhatConfig, HardhatRuntimeEnvironment, HardhatUserConfig, HttpNetworkConfig } from "hardhat/types";
 import { adaptLog, adaptUrl, getDefaultHttpNetworkConfig } from "./utils";
 import { DockerWrapper, VenvWrapper } from "./starknet-wrappers";
+import { glob } from "hardhat/internal/util/glob";
```
FabijanC, line 13: `glob` should be directly imported, not depending on it being in hardhat internals.

src/index.ts
```diff
@@ -10,31 +10,8 @@ import { PLUGIN_NAME, ABI_SUFFIX, DEFAULT_STARKNET_SOURCES_PATH, DEFAULT_STARKNE
...
-    if (stats.isDirectory()) {
-        for (const childName of fs.readdirSync(traversable)) {
-            const childPath = path.join(traversable, childName);
-            statusCode += await traverseFiles(childPath, predicate, action);
-        }
-    } else if (stats.isFile()) {
-        if (predicate(traversable)) {
-            statusCode += await action(traversable);
-        }
-    } else {
```
FabijanC, line 31: As we discussed, these three cases should be handled (dir, file, other), the third case not being supported.

FabijanC: To avoid duplicating this checking scheme, you could have a single function doing that (and the function could also accept the later used filter criterion).

FabijanC (changes requested):

src/index.ts
FabijanC: After the call, we mentioned potential use of `util.promisify ` or something like that.

src/index.ts
```diff
@@ -248,6 +246,19 @@ task("starknet-compile", "Compiles Starknet contracts")
...
+async function traverseFiles(traversable: string, fileCriteria: string = "*") {
+    let paths: string[] = [];
+    if (fs.lstatSync(traversable).isDirectory()) {
+        const globPromise = promisify(glob);
+        paths = await globPromise(path.join(traversable, "**", fileCriteria));
+    }
+    else {
+        paths.push(traversable);
+    }
+    const files = paths.filter(file => { return fs.lstatSync(file).isFile(); });
```
FabijanC, line 258: no need for using the `return` statement and curly brackets, this should be valid: `file => fs.lstatSync(file).isFile()`

src/index.ts
```diff
@@ -373,22 +378,19 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
...
-    
-    return foundPath;
+    if(files.length == 0){
+        return null;
+    }
+    else if(files.length == 1){
+        return files[0];
+    }
+    else {
+        throw new HardhatPluginError(PLUGIN_NAME, `More than one file was found because the path provided is ambiguous, please specify a relative path`);
```
FabijanC, line 392: This is a rather long line. Handle it as in other such places of this project. Also no need for the template literal (the backtick string definition) if nothing is interpolated with `${}`)

src/index.ts
```diff
@@ -248,6 +246,19 @@ task("starknet-compile", "Compiles Starknet contracts")
         }
     });
 
+async function traverseFiles(traversable: string, fileCriteria: string = "*") {
+    let paths: string[] = [];
+    if (fs.lstatSync(traversable).isDirectory()) {
+        const globPromise = promisify(glob);
```
FabijanC, line 252: Instead of promisifying this on each `traverseFiles` call, I'd suggest creating a global const function right after the imports at the top of the file.

src/index.ts
```diff
@@ -34,6 +35,10 @@ function checkArtifactExists(artifactsPath: string): void {
  * @returns 0 if succeeded, 1 otherwise
  */
 function processExecuted(executed: ProcessResult): number {
+    if(!executed.stdout){
+        const msg = "Command does not exist. Check if 'cairo-lang' is installed in your active environment.";
```
FabijanC, line 39: As we said in the call, we aren't sure if this is the only case where `executed.stdout` might be falsy. The message could be framed as:
`Command not found. If you're using a Python virtual environment, check that it has 'cairo-lang' installed.`

src/index.ts
FabijanC: Also, does providing relative path work? We should have it tested in starknet-hardhat-example

FabijanC (changes requested):
Also as discussed, Readme should have the paths issue clarified (PATH.cairo).

FabijanC (approved):

Tiago-shardlabs (approved):

PR #34, opened 2021-12-20
Added short string support, refactored index.ts

dribeiro-ShardLabs:


FabijanC (changes requested):

FabijanC (approved):
The documentation of string2bigint will probably be refined in future, but I'd say this PR is ready, so I'll merge it.

PR #36, opened 2021-12-27
Added check for number of arguments and nested objects' members

dribeiro-ShardLabs:
Considering the cairo function:
﻿```func foo(a: felt) -> (res: felt):
﻿end```﻿

The following testing code samples with incorrect number of arguments should fail
```
contract.call("foo", { a: 1, b: 2 });
contract.call("foo"); 
```

In functions that take arrays as arguments, the array length is not considered as an argument.

### DIFF
--- a/src/adapt.ts
+++ b/src/adapt.ts
@@ -134,8 +134,8 @@ function adaptComplexInput(input: any, inputSpec: starknet.Argument, abi: starkn
 
         const memberTypes = type.slice(1, -1).split(", ");
 
-        if (input.length != 2) {
-            const msg = `Expected "${inputSpec.name}" to have exactly 2 arguments`;
+        if (input.length != memberTypes.length) {
+            const msg = `"${inputSpec.name}": Expected ${memberTypes.length} member${memberTypes.length === 1 ? "" : "s"}, got ${input.length}.`;
             throw new HardhatPluginError(PLUGIN_NAME, msg);
         }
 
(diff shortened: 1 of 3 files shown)


src/adapt.ts
```diff
@@ -41,6 +41,22 @@ export function adaptInput(functionName: string, input: any, inputSpecs: starkne
     const adapted: string[] = [];
     let lastSpec: starknet.Argument = { type: null, name: null };
 
+    // User won't pass array length as an argument, so subtract the number of array elements to the expected amount of arguments
+    const countArrays = inputSpecs.filter( i => i.name.endsWith(LEN_SUFFIX)).length;
+    const expectedInputCount = inputSpecs.length-countArrays;
+
+    if(expectedInputCount > 0 && !input) {
```
FabijanC, line 48: Can't this be rewritten using just one if ?

src/adapt.ts
```diff
@@ -41,6 +41,22 @@ export function adaptInput(functionName: string, input: any, inputSpecs: starkne
     const adapted: string[] = [];
     let lastSpec: starknet.Argument = { type: null, name: null };
 
+    // User won't pass array length as an argument, so subtract the number of array elements to the expected amount of arguments
+    const countArrays = inputSpecs.filter( i => i.name.endsWith(LEN_SUFFIX)).length;
```
FabijanC, line 45: This will not always be correct, e.g. in case an input argument has the len suffix, but does not precede an actual array. A more complete check can be found in this same function (line 68). Also I think the current implementation allows passing the len argument, although not as a required argument.

FabijanC (changes requested):
Extra arguments (object keys) can appear in nested objects; that should be handled as well.

src/adapt.ts
```diff
@@ -122,6 +134,11 @@ function adaptComplexInput(input: any, inputSpec: starknet.Argument, abi: starkn
 
         const memberTypes = type.slice(1, -1).split(", ");
 
+        if (input.length != 2) {
```
FabijanC, line 137: Why 2?

FabijanC (approved):

PR #38, opened 2021-12-31
Log available networks

FabijanC:
Description Pending

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@ import * as path from "path";
 import { task, extendEnvironment, extendConfig } from "hardhat/config";
 import { HardhatPluginError } from "hardhat/plugins";
 import "./type-extensions";
-import { PLUGIN_NAME, DEFAULT_STARKNET_SOURCES_PATH, DEFAULT_STARKNET_ARTIFACTS_PATH, DEFAULT_DOCKER_IMAGE_TAG, DOCKER_REPOSITORY, ALPHA_URL, ALPHA_MAINNET_URL } from "./constants";
+import { PLUGIN_NAME, DEFAULT_STARKNET_SOURCES_PATH, DEFAULT_STARKNET_ARTIFACTS_PATH, DEFAULT_DOCKER_IMAGE_TAG, DOCKER_REPOSITORY, ALPHA_URL, ALPHA_MAINNET_URL, VOYAGER_GOERLI_CONTRACT_API_URL, VOYAGER_MAINNET_CONTRACT_API_URL } from "./constants";
 import { HardhatConfig, HardhatUserConfig } from "hardhat/types";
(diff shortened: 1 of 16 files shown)


src/index.ts
```diff
@@ -60,11 +60,11 @@ extendConfig((config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) =>
 // add url to alpha network
 extendConfig((config: HardhatConfig) => {
```
dribeiro-ShardLabs, line 61: Maybe it makes sense to add a default verification URL if the user does not provide one, or at least return an error.

test/configuration-tests/with-networks/check.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Testing no mocha network"
+NETWORK=""    npx hardhat test test/contract-factory-test.ts
+echo "Success"
+
+echo "Testing invalid mocha network"
+NETWORK="$INVALID_NETWORK" npx hardhat test test/contract-factory-test.ts 2>&1 \
+    | tail -n +9 \
+    | head -n 6 \
+    | diff - "$PREFIX/invalid-mocha-network.txt"
+echo "Success"
```
dribeiro-ShardLabs, line 30: We should also add a test for the verification and invoke tasks

src/task-actions.ts
```diff
@@ -224,21 +207,15 @@ export async function starknetDeployAction(args: any, hre: HardhatRuntimeEnviron
...
-    if (!isTestnet(args.starknetNetwork)) {
-        if (isMainnet(args.starknetNetwork)) {
-            voyagerUrl = VOYAGER_MAINNET_CONTRACT_API_URL;
-        } else {
-            const msg = `Unknown starknet-network provided: ${args.starknetNetwork}`;
-            throw new HardhatPluginError(PLUGIN_NAME, msg);
-        }
+    const network = getNetwork(args.starknetNetwork, hre, "starknet-network");
+    if (!network.verificationUrl) {
+        throw new HardhatPluginError(PLUGIN_NAME, `Network ${args.starknetNetwork} does not support Voyager verification.`);
```
dribeiro-ShardLabs, line 212: Is there any check for the verification URL? It should be only either testnet or mainnet

dribeiro-ShardLabs (commented):
There should be a verification for the verificationURL to be either testnet or mainnet (networks supported by voyager at the moment)

dribeiro-ShardLabs (approved):

test/configuration-tests/with-networks/check.sh
FabijanC: The verification tasks are temporarily disabled until (we are blocked by the Voyager team on that), and invoke tasks aren't supported yet, but might be in future.

PR #39, opened 2022-01-11
Invoke call tasks

### LINKED ISSUE #40: `starknet-deploy` should support `salt` parameter
Python CLI deploy command supports `salt` parameter:
```
(cairo_venv) maciejka@X1$ starknet deploy
usage: starknet [-h] [--salt SALT] [--inputs [INPUTS [INPUTS ...]]] --contract
                CONTRACT [--token TOKEN]
starknet: error: the following arguments are required: --contract
```
hardhat plug in does not seem to do it: https://github.com/Shard-Labs/starknet-hardhat-plugin/blob/1bd7c2a34198155e95137849fcf36536c296d4e6/src/starknet-wrappers.ts#L17

`Deploy` method of `StarknetContractFactory` (https://github.com/Shard-Labs/starknet-hardhat-plugin/blob/1bd7c2a34198155e95137849fcf36536c296d4e6/src/types.ts#L163) should also support this parameter.

dribeiro-ShardLabs:
Added `starknet-invoke` and `starknet-call` plugin tasks.

Changed the `starknet-deploy` task so it also accepts contract names instead of just paths.

Added a new optional parameter on `getContractFactoryUtil`, `networkURL?:string`
If that argument is provided, it is used instead of the network in the hardhat config file.

Added salt parameter to `starknet-deploy`

fix/close #40


src/extend-utils.ts
```diff
@@ -3,25 +3,12 @@ import { HardhatPluginError } from "hardhat/plugins";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { ABI_SUFFIX, DEFAULT_STARKNET_NETWORK, PLUGIN_NAME, SHORT_STRING_MAX_CHARACTERS } from "./constants";
 import { StarknetContractFactory } from "./types";
-import { checkArtifactExists, traverseFiles, getNetwork } from "./utils";
+import { checkArtifactExists, findPath, getNetwork } from "./utils";
```
FabijanC, line 6: No need to have four newlines after imports.

src/extend-utils.ts
```diff
@@ -39,17 +26,23 @@ export async function getContractFactoryUtil (hre: HardhatRuntimeEnvironment, co
         throw new HardhatPluginError(PLUGIN_NAME, `Could not find ABI for ${contractPath}`);
     }
 
+    let gateway;
```
FabijanC, line 29: This variable should be declared as late as possible.

src/extend-utils.ts
```diff
@@ -39,17 +26,23 @@ export async function getContractFactoryUtil (hre: HardhatRuntimeEnvironment, co
...
     const testNetworkName = hre.config.mocha.starknetNetwork || DEFAULT_STARKNET_NETWORK;
-
-    const network = getNetwork(testNetworkName, hre, "mocha.starknetNetwork");
     hre.starknet.network = testNetworkName;
```
FabijanC, line 31: Does it make sense to set `hre.starknet.network` if it's not the one being used. This is possible in case networkURL is provided. Would it make more sense to have networkName as an argument instead of networkURL? Or the hre.starknet.network setting could be done outside this function?

test/configuration-tests/with-deploy-path/check.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+npx hardhat starknet-compile contracts/contract.cairo
+npx hardhat starknet-deploy starknet-artifacts/contracts/contract.cairo/ --inputs "10" --starknet-network "$NETWORK"
+npx hardhat starknet-deploy contract --inputs "10" --starknet-network "$NETWORK"
```
FabijanC, line 6: newline

test/configuration-tests/with-deploy-inputs/check.sh
```diff
@@ -3,3 +3,4 @@ set -e
 
 npx hardhat starknet-compile contracts/contract.cairo
 npx hardhat starknet-deploy starknet-artifacts/contracts/contract.cairo/ --inputs "10" --starknet-network "$NETWORK"
+npx hardhat starknet-deploy starknet-artifacts/contracts/contract.cairo/ --inputs "10" --starknet-network "$NETWORK" --salt 0x10
```
FabijanC, line 6: newline

src/starknet-wrappers.ts
```diff
@@ -65,6 +66,10 @@ export abstract class StarknetWrapper {
             prepared.push("--signature", ...options.signature);
         }
 
+        if (options.salt && options.salt !== "") {
```
FabijanC, line 69: Why the extra check for `!== ""`?

src/task-actions.ts
```diff
@@ -277,3 +282,37 @@ export async function starknetVoyagerAction(args: any, hre: HardhatRuntimeEnviro
...
+export async function starknetInvokeAction(args: any, hre: HardhatRuntimeEnvironment) {
+    await starknetInvokeOrCallAction("invoke", args, hre);
+}
+
+export async function starknetCallAction(args: any, hre: HardhatRuntimeEnvironment) {
+    await starknetInvokeOrCallAction("call", args, hre);
+}
+
+
+async function starknetInvokeOrCallAction(choice: any, args: any, hre: HardhatRuntimeEnvironment) {
```
FabijanC, line 295: No need for choice to be `any`. Maybe Enum with string values can be used.

src/index.ts
```diff
@@ -106,7 +106,14 @@ task("starknet-deploy", "Deploys Starknet contracts which have been compiled.")
         "Pass them as a single string; e.g. --inputs \"1 2 3\"\n" +
         "You would typically use this feature when deploying a single contract.\n" +
         "If you're deploying multiple contracts, they'll all use the same input."
-    ).addOptionalVariadicPositionalParam("paths",
+    )
+    .addOptionalParam("salt",
+        "An optional salt controlling where the contract will be deployed.\n" +
+        "The contract deployment address is determined by the hash\n" +
```
FabijanC, line 112: Why the newline?

src/task-actions.ts
```diff
@@ -156,10 +156,14 @@ export async function starknetDeployAction(args: any, hre: HardhatRuntimeEnviron
     let statusCode = 0;
     const txHashes: string[] = [];
     for (let artifactsPath of artifactsPaths) {
-        if (!path.isAbsolute(artifactsPath)) {
+
+        // Check if input is the name of the contract and not a path
+        if (artifactsPath === path.basename(artifactsPath)) {
```
FabijanC, line 161: What does this introduce? When deploying, contract name (with extension) can be provided? Is that the only case supported with the addition of these lines? And is this tested?

FabijanC (changes requested):
StarknetContractFactory should also contain salt option.

FabijanC (approved):
Looks good, just wondering why the last commit is a merge commit?

PR #41, opened 2022-01-14
Using caching in CircleCI to optimize workflow.

dribeiro-ShardLabs:
Optimized CircleCI workflow to use caching


.circleci/config.yml
```diff
@@ -9,18 +9,109 @@ executors:
...
+  - &save_devnet_cache
+    - save_cache:
+          name: "Save devnet Cache"
+          key: devnet-cache-{{ arch }}-{{ checksum "/tmp/devnet-version" }}
+          paths:
+            - /opt/circleci/.pyenv/versions
+
+commands:
+  fetch_dependecies_version:
+    description: "Fetch the dendencies version"
```
FabijanC, line 67: dendencies

.circleci/config.yml
```diff
@@ -9,18 +9,109 @@ executors:
       xcode: 11.7
     environment:
       TERM: xterm-256color
-    
+
+aliases:
+  - &restore_macos_brew_cache
```
FabijanC, line 14: why is this an alias? it's not reused, nor is any of its keys assigned when being referenced (<<: *). Perhaps the alias part could be the `when` condition, and the steps could be specified on reference.

.circleci/config.yml
```diff
@@ -42,15 +135,14 @@ jobs:
...
 
-
 workflows:
   test_and_publish:
     jobs:
       - test:
           matrix:
             parameters:
               platform: [linux-node14, macos-node14]
-              test-name: [test-general-tests, test-configuration-tests, test-venv-tests]
```
FabijanC, line 53: Why is the order changed?

scripts/test.sh
```diff
@@ -56,7 +56,10 @@ if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$OSTYPE" == "linux-gnu"* ]]; then
 fi
 
 # install and build devnet
-../scripts/install-devnet.sh
+
+export PATH="$PATH:/opt/circleci/.pyenv/shims"
```
FabijanC, line 60: Why this PATH manipulation?

.circleci/config.yml
```diff
@@ -9,18 +9,109 @@ executors:
...
+                    key: brew-gmp-cache-{{ arch }}-v1
+                    paths:
+                      - /usr/local/Cellar/gmp
+                      - /usr/local/lib
+                      - /usr/local/include
+
+  - &restore_cairo_cache
+    - restore_cache:
+          name: "Restore cairo-lang Cache"
+          key: cairo-lang-cache-{{ arch }}-{{ checksum "/tmp/cairo-lang-version" }}
```
FabijanC, line 40: I've given this some more thought. There's a danger in doing this in case plugin's not yet been adapted to work with the latest version. This is clear if we observe the setup-venv script which has the plugin version hardcoded.

The ideal approach would be to have the current cairo-lang to which the plugin is adapted only in one place, but so far we haven't been able to come up with that.

Probably we'll have to hardcode it here as well. That way we'll have two more places (restore and save) to change when adapting to new cairo-lang versions (there are a few already across the project, e.g. Readme). Or in case we can utilize variables config.yml, there'd only be one more place.

FabijanC (changes requested):

FabijanC (approved):

PR #43, opened 2022-01-19
Implemented support for pending block querying

dribeiro-ShardLabs:
Add support for pending block


src/index.ts
```diff
@@ -173,6 +173,7 @@ task("starknet-call", "Invokes a function on a contract in the provided address.
         "Pass them as a single string; e.g. --inputs \"1 2 3\"")
     .addOptionalParam("signature", "The call signature")
     .addOptionalParam("wallet", "The wallet to use, defined in the 'hardhat.config' file. If omitted, the '--no_wallet' flag will be passed when calling.")
+    .addOptionalParam("blockNumber", "The number of the block to call")
```
FabijanC, line 176: Would it make sense to note that "pending" is also acceptable: "The number of the block to call or \"pending\""? Or maybe even the PENDING_BLOCK_NUMBER const could be used.

FabijanC (approved):

PR #44, opened 2022-01-20
Parse negative hexadecimal outputs to BigInt

clexmond:
StarkNet calls will return negative numbers (ex. `-0x001`) but `BigInt` can't parse negative hexadecimal strings (`BigInt("-0x001")` will fail with `Cannot convert -0x001 to a BigInt`). `BigInt` will happily parse both `-0x001` (not as a string) and `"0x001"` (positive value as a string).

This change simply strips the sign off, converts the (now positive) hexadecimal value to `BigInt` and negates before returning.


dribeiro-ShardLabs (approved):
Looks good :) thank you for the contribution

PR #47, opened 2022-01-26
Implement support for account contract

dribeiro-ShardLabs:
- Implemented task `starknet-deploy-account` ;
- Modified tasks `starknet-invoke` and `starknet-call` to accept a wallet instead of a signature;
- Added method `getWallet` to Starknet object;
- Extended types to support wallet object in `hardhat.config` file;
- Modified methods `invoke` and `call` in StarknetContract to accept a wallet;


src/extend-utils.ts
```diff
@@ -75,3 +77,13 @@ export function bigIntToStringUtil(convertableBigInt: BigInt) {
 }
 
 
+export function getWalletUtil(name: string, hre: HardhatRuntimeEnvironment) {
+    const wallet = hre.config.wallets[name];
+    if (!wallet) {
+        const available = Object.keys(hre.config.wallets).join(", ");
+        const msg = `Invalid wallet provided: ${name}.\nValid wallets: ${available}`;
```
FabijanC, line 84: The wallet name is what's invalid, not the wallet itself.

src/task-actions.ts
```diff
@@ -301,14 +301,29 @@ async function starknetInvokeOrCallAction(choice: Choice, args: any, hre: Hardha
     const contractFactory = await hre.starknet.getContractFactory(args.contract, gatewayUrl);
     const abiPath = contractFactory.getAbiPath();
 
+    let wallet, accountDir;
+    if (args.wallet) {
+        wallet = hre.config.wallets[args.wallet];
+        if (!wallet) {
```
FabijanC, line 307: Can lines 304-313 be replaced/merged with `getWalletUtil`?

src/task-actions.ts
```diff
@@ -320,3 +335,33 @@ async function starknetInvokeOrCallAction(choice: Choice, args: any, hre: Hardha
...
     }
 }
+
+export async function starknetDeployAccountAction(args: any, hre: HardhatRuntimeEnvironment) {
+    const gatewayUrl = getGatewayUrl(args, hre);
+    const wallet = hre.config.wallets[args.wallet];
+
+    if (!wallet) {
+        const available = Object.keys(hre.config.wallets).join(", ");
+        const msg = `Invalid wallet provided: ${args.wallet}.\nValid wallets: ${available}`;
```
FabijanC, line 345: Can lines 341-349 be replaced/merged with `getWalletUtil`?

src/starknet-wrappers.ts
```diff
@@ -199,6 +248,10 @@ export class DockerWrapper extends StarknetWrapper {
             [options.abi]: options.abi
         };
 
+        if (options.accountDir) {
+            addPaths(binds, options.accountDir);
```
FabijanC, line 252: Why not just directly populate the `binds` object? The `addPaths` function is for the cairo-path option of the compile task (I should've better commented that one).

src/starknet-wrappers.ts
```diff
@@ -103,6 +132,25 @@ export abstract class StarknetWrapper {
     }
 
     public abstract getTxStatus(options: GetTxStatusOptions): Promise<ProcessResult>;
+
+    protected prepareDeployAccountOptions(options: DeployAccountOptions): string {
+        const wallet = options.wallet? "'" + options.wallet + "'" : "None";
+        const accountName = options.accountName? "'" + options.accountName + "'" : "'__default__'";
+        const accountDir = options.accountDir? "'" + options.accountDir + "'" : "None";
+        return "Namespace(" +
```
FabijanC, line 140: I'm almost sure this can be written in a better way, we can discuss it. Basically the idea of `prepareXOptions` was to prepare CLI options for the StarknetCLI commands that are being wrapped.

README.md
```diff
@@ -295,5 +310,34 @@ module.exports = {
...
+You can specify multiple wallets/accounts.
+
+The parameters for the wallet are:
+  - `accountName`: The name to give the account. If omitted, the default value `__default__ ` will be used;
+  - `modulePath`: The python module and wallet class of your chosen wallet provider;
+  - `accountPath`: The path where your wallet information will be saved.
+
+```javascript
+module.exports = {
+  ...
```
FabijanC, line 324: I remember putting the ellipsis before and after blocks when showing example usage, but ended up having only the one after the block.

FabijanC (changes requested):
Check the linter output for unused imports (is it just a warning?, I thought unused imports would give linting error).

src/extend-utils.ts
dribeiro-ShardLabs: Here it means that if the wallet name is not in the hardhat.config file, the wallet with that name doesent exist, so "wallet" will be undefined

FabijanC (approved):

PR #48, opened 2022-01-27
Multiple sources paths, rejected transaction error message

dribeiro-ShardLabs:
- Multiple paths are allowed now for `hardhat.config` file.
- Error message from a rejected transaction now displayed.

### DIFF
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -117,7 +117,7 @@ export async function starknetCompileAction(args: any, hre: HardhatRuntimeEnviro
         }
         checkSourceExists(sourcesPath);
         const files = await traverseFiles(sourcesPath, "*.cairo");
-        const defaultSourcesPaths = defaultSourcesPath.replace(" ", ":");
+        const defaultSourcesPaths = defaultSourcesPath.replace(/\s+/g, " ").replace(" ", ":");
         for (const file of files) {
             console.log("Compiling", file);
             const suffix = file.replace(rootRegex, "");
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -332,6 +332,26 @@ async function starknetInvokeOrCallAction(choice: Choice, args: any, hre: Hardha
         const replacedMsg = adaptLog(msg);
         throw new HardhatPluginError(PLUGIN_NAME, replacedMsg);
     }
+
+    if (choice === "invoke" && args.wait) { // If the "wait" flag was passed as an argument, check the transaction hash for its status
+        console.log(`Checking ${choice} transaction...`);
+        const executedOutput = executed.stdout.toString();
+        const txHash = extractTxHash(executedOutput);
+        await new Promise<void>((resolve, reject) => iterativelyCheckStatus(
+            txHash,
+            hre.starknetWrapper,
+            gatewayUrl,
+            gatewayUrl,
+            status => {
+                console.log(`Invoke transaction ${txHash} is now ${status}`);
+                resolve();
+            },
+            error => {
+                console.log(`Invoke transaction ${txHash} is REJECTED`);
+                reject(error);
+            }
+        ));
+    }
 }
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -362,7 +369,10 @@ export class StarknetContract {
                 this.gatewayUrl,
                 this.feederGatewayUrl,
                 () => resolve(),
-                reject
+                error => {
+                    console.log(`Invoke transaction ${txHash} is REJECTED.\n` + error.message);
```
FabijanC, line 373: Would console.error make more sense?

src/types.ts
```diff
@@ -52,6 +58,14 @@ export interface StringMap {
 
 export type Choice = "call" | "invoke";
 
+export function extractTxHash(response: string) {
+    return extractFromResponse(response, /^Transaction hash: (.*)$/m);
+}
+
+export function extractAddress(response: string) {
```
FabijanC, line 65: Does this need to be exported? And why were these two extract functions moved up a few lines?

src/task-actions.ts
```diff
@@ -319,4 +319,24 @@ async function starknetInvokeOrCallAction(choice: Choice, args: any, hre: Hardha
...
+            txHash,
+            hre.starknetWrapper,
+            gatewayUrl,
+            gatewayUrl,
+            status => {
+                console.log(`Invoke transaction ${txHash} is now ${status}`);
+                resolve();
+            },
+            error => {
+                console.log(`Invoke transaction ${txHash} is REJECTED`);
```
FabijanC, line 337: Would it make sense to use console.error here?

src/task-actions.ts
```diff
@@ -106,25 +106,25 @@ export async function starknetCompileAction(args: any, hre: HardhatRuntimeEnviro
...
 
     let statusCode = 0;
     for (let sourcesPath of sourcesPaths) {
         if (!path.isAbsolute(sourcesPath)) {
             sourcesPath = path.normalize(path.join(root, sourcesPath));
         }
-
         checkSourceExists(sourcesPath);
         const files = await traverseFiles(sourcesPath, "*.cairo");
+        const defaultSourcesPaths = defaultSourcesPath.replace(" ", ":");
```
FabijanC, line 119: What if `defaultSourcesPath` contains multiple consecutive spaces? And where is the input format specified/documented?

FabijanC (changes requested):

src/task-actions.ts
```diff
@@ -117,7 +117,7 @@ export async function starknetCompileAction(args: any, hre: HardhatRuntimeEnviro
         }
         checkSourceExists(sourcesPath);
         const files = await traverseFiles(sourcesPath, "*.cairo");
-        const defaultSourcesPaths = defaultSourcesPath.replace(" ", ":");
+        const defaultSourcesPaths = defaultSourcesPath.replace(/\s+/g, " ").replace(" ", ":");
```
FabijanC, line 120: Why not just one replacement directly to ":"?

FabijanC (changes requested):

src/task-actions.ts
FabijanC: This remained, do we keep this as it is?

src/task-actions.ts
```diff
@@ -110,12 +110,29 @@ export async function starknetCompileAction(args: any, hre: HardhatRuntimeEnviro
     const sourcesPaths: string[] = args.paths || [defaultSourcesPath];
     const artifactsPath = hre.config.paths.starknetArtifacts;
 
+    const cairoPaths = [defaultSourcesPath, root];
+    if (args.cairoPath) {
+        args.cairoPath.split(":").forEach((path: string) => {
+            cairoPaths.push(path);
+        });
+    }
+    if (hre.config.paths.cairoPaths) {
```
FabijanC, line 119: This should be documented in Readme

FabijanC (approved):

PR #57, opened 2022-02-16
Updated to cairo 0.7.1

dribeiro-ShardLabs:
Update to cairo version 0.7.1

Readme will be updated when updating the semver and creating the release tag


FabijanC (changes requested):
Also update Readme. Or if you can think of a way so that Readme dosn't require manual update, that's also welcome.

dribeiro-ShardLabs:
> Also update Readme. Or if you can think of a way so that Readme dosn't require manual update, that's also welcome.

Will update the Readme when updating the semver and creating the release tag

FabijanC (approved):

PR #58, opened 2022-02-16
Postman

badurinantun:
### Usage

* `starknet.devnet` exposes `devnet` specific utility methods for the communication between L1 and L2
* `starknet.devnet.flush` method sends pending messages
* `starknet.devnet.loadL1MessagingContract` method deploys or loads the mock messaging contract
* `starknet.devnet` methods should throw an error if the network doesn't have `/postman` endpoint


src/type-extensions.ts
```diff
@@ -56,6 +56,40 @@ type StarknetContractFactoryType = StarknetContractFactory;
 type StringMapType = StringMap;
 
 declare module "hardhat/types/runtime" {
+    interface L2Message {
```
dribeiro-ShardLabs, line 59: I think these interfaces could be along with the DevUtils class in a different file, since they're only used by DevUtils

src/extend-utils.ts
```diff
@@ -86,3 +88,42 @@ export function getWalletUtil(name: string, hre: HardhatRuntimeEnvironment) {
     wallet.accountPath = getAccountPath(wallet.accountPath, hre);
     return wallet;
 }
+
```
dribeiro-ShardLabs, line 91: I think DevnetUtils could be in a different file, since it has its own purpose

badurinantun: [da02ed0](https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/58/commits/da02ed0b8e4dd3daa4ef5c8b2783dd7aafdc6272)

src/devnet-utils.ts
```diff
@@ -0,0 +1,62 @@
...
+import { HardhatPluginError } from "hardhat/plugins";
+import { Devnet, HardhatRuntimeEnvironment } from "hardhat/types";
+
+import { PLUGIN_NAME } from "./constants";
+import { getNetwork } from "./utils";
+
+interface L2Message {
+  from_address: BigInt;
+  to_address: BigInt;
+  payload: Array<any>;
```
maciejka, line 12: Why not Array<BigInt>?

test/general-tests/postman/check.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+trap 'kill $(jobs -p)' EXIT
+
+npx hardhat starknet-compile
+npx hardhat compile
+npx hardhat node &
+sleep 1
+
+npx hardhat test --network localhost test/postman.test.ts
```
maciejka, line 11: `test/postman.test.ts` does not seem to be added to the commit?

src/devnet-utils.ts
badurinantun: Good call, changed it ab5baf3

scripts/test.sh
```diff
@@ -31,6 +31,8 @@ function iterate_dir(){
         test_name=$(basename $test_case)
         if [[ "$network" == "devnet" ]] && [[ "$test_name" == "wallet-test" ]]; then
             echo "Skipping devnet test for wallet-test"
+        elif [[ "$network" == "alpha"]] && [[ "$test_name" == "postman"]]; then
```
FabijanC, line 34: Shouldn't there be a space before `]]`

scripts/test.sh
```diff
@@ -31,6 +31,8 @@ function iterate_dir(){
         test_name=$(basename $test_case)
         if [[ "$network" == "devnet" ]] && [[ "$test_name" == "wallet-test" ]]; then
             echo "Skipping devnet test for wallet-test"
+        elif [[ "$network" == "alpha"]] && [[ "$test_name" == "postman" ]]; then
```
FabijanC, line 34: missing space after "alpha"

dribeiro-ShardLabs (approved):

PR #59, opened 2022-02-16
Move hardhat to peerDependencies

badurinantun:
Moved `hardhat` to `peerDependencies`


FabijanC (changes requested):
https://hardhat.org/advanced/building-plugins.html#notes-on-dependencies
Rule of thumb: Every peerDependency should also be a devDependency.

dribeiro-ShardLabs (changes requested):
Pending Fabijan's change requests

dribeiro-ShardLabs (approved):

PR #62, opened 2022-02-21
Account enhancement

dribeiro-ShardLabs:
Add Account class, which serves as a proxy for contract interaction
- Removed open-zeppelin-cairo-contracts test, which is no longer needed


ccarnino (commented):
It looks like a real step forward in the account management!

src/index.ts
```diff
@@ -175,7 +177,23 @@ extendEnvironment((hre) => {
             return wallet;
         },
 
-        devnet: lazyObject(() => new DevnetUtils(hre))
+        devnet: lazyObject(() => new DevnetUtils(hre)),
+
+        deployAccountFromABI: async (accountContract, accountType) => {
```
FabijanC, line 182: Will not be needed if we decide to expose OpenZeppelinAccount class with its constructor.

src/type-extensions.ts
```diff
@@ -126,11 +127,38 @@ declare module "hardhat/types/runtime" {
...
 
             devnet: Devnet;
+
+            /**
+             * Deploys an Account contract based on the ABI and the type of Account selected
+             * @param accountContract the case-sensitive contract name, same as {@link getContractFactory}
+             * @param accountType the enumerator value of the type of Account to use
+             * @returns an Account object
+             */
+            deployAccountFromABI: (
```
FabijanC, line 137: Will not be needed if we decide to expose OpenZeppelinAccount class with its constructor.

src/extend-utils.ts
```diff
@@ -104,3 +105,44 @@ export function getWalletUtil(name: string, hre: HardhatRuntimeEnvironment) {
...
+
+export async function deployAccountFromABIUtil(
+    accountContract: string,
+    accountType: AccountTypes,
+    hre: HardhatRuntimeEnvironment
+): Promise<Account> {
+    let account: Account;
+    switch (accountType) {
+        case "OpenZeppelin":
+            account = await OpenZeppelinAccount.deployFromABI(accountContract, hre);
```
FabijanC, line 117: Will not be needed if we decide to expose OpenZeppelinAccount class with its constructor.

src/account.ts
```diff
@@ -0,0 +1,231 @@
...
+
+        const pk = "0x" + privateKey.toString(16);
+        console.log("Account private key: " + pk);
+        console.log("Account public key: " + publicKey);
+        console.log("Account address: " + contract.address);
+
+        return new OpenZeppelinAccount(contract, pk, publicKey, keyPair);
+    }
+
+    static async getAccountFromAddress(
```
FabijanC, line 195: Instead of having static functions for each Account implementation and starknet utility functions calling them, we could do this:
- The constructor could take parameters, potentially even an options object, and generate/retrieve what it doesn't have.
- We wouldn't need the switch-case logic which would require future updates with new Account implementations.

src/account.ts
```diff
@@ -0,0 +1,231 @@
...
+     * Adapts the input arguments to the proper format to use in the Account contract proxy invocation function that uses the call_contract syscall
+     *
+     * @param toAddress address of the contract to be called
+     * @param selector function in the contract to be called
+     * @param calldata calldata to use as input for the contract call
+     */
+    abstract adaptArgs(toAddress: string, selector: string, calldata: StringMap): any;
+
+    /**
+     * Recursively transforms the input data received as an object to a <Numeric> array
```
FabijanC, line 69: Numeric or BigNumberish?

FabijanC (changes requested):

src/utils.ts
```diff
@@ -158,3 +159,17 @@ export function getAccountPath(accountPath: string, hre: HardhatRuntimeEnvironme
     }
     return accountDir;
 }
+
+export async function getCodeFromAddress(address: string, networkURL: string) {
+    const getCodeURL = networkURL + "/feeder_gateway/get_code";
```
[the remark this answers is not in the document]
dribeiro-ShardLabs: Method will be deleted because it's actually not needed anymore

src/constants.ts
```diff
@@ -22,5 +22,6 @@ export const VOYAGER_GOERLI_CONTRACT_API_URL = "https://goerli.voyager.online/ap
 export const VOYAGER_MAINNET_CONTRACT_API_URL = "https://voyager.online/api/contract/";
 
 export const SHORT_STRING_MAX_CHARACTERS = 31;
+export const OPENZEPPELIN_EXECUTE_FUNCTION = "execute";
```
FabijanC, line 25: Better to have it as a `static readonly` string on OpenZeppelinAccount clas itself.

FabijanC (changes requested):

src/account.ts
FabijanC: Dropping this idea, no dev-friendly and user-friendly way of implementing it.

FabijanC (approved):
Readme corrections can be done with the release of a new version. Can be merged.

PR #68, opened 2022-03-02
Updated starknet-verify action

dribeiro-ShardLabs:
Adapt the contract verification to be compatible with the new Voyager update

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -201,7 +201,13 @@ extendEnvironment((hre) => {
 
 task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
-    .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
+    .addOptionalParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
+    .addOptionalParam(
+        "paths",
+        "The paths of the cairo contract and its internal dependencies, in this exact order, separated by a space." +
+            "All dependencies should be in the same folder as the contract." +
+            "e.g. --paths='path/to/main path/to/dependency1 path/to/dependency2'"
+    )
     .addParam("address", "The address where the contract is deployed")
     .setAction(starknetVoyagerAction);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -201,11 +201,11 @@ extendEnvironment((hre) => {
 
 task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
-    .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
+    .addParam("path", "The path of the main cairo contract (e.g. contracts/contract.cairo)")
     .addParam("address", "The address where the contract is deployed")
     .addOptionalVariadicPositionalParam(
     "paths",
-    "The paths of the dependencies of the contract specified in --param" +
+    "The paths of the dependencies of the contract specified in --path" +
             "All dependencies should be in the same folder as the contract." +
             "e.g. path/to/dependency1 path/to/dependency2"
     )
--- a/src/index.ts
+++ b/src/index.ts
@@ -201,14 +201,14 @@ extendEnvironment((hre) => {
 
 task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
-    .addOptionalParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
-    .addOptionalParam(
-        "paths",
-        "The paths of the cairo contract and its internal dependencies, in this exact order, separated by a space." +
-            "All dependencies should be in the same folder as the contract." +
-            "e.g. --paths='path/to/main path/to/dependency1 path/to/dependency2'"
-    )
+    .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -201,7 +201,13 @@ extendEnvironment((hre) => {
 
 task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
-    .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
+    .addOptionalParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
+    .addOptionalParam(
```
FabijanC, line 205: Why have both `path` and `paths`? Shouldn't it be just `path` which is a variadic parameter? Also that way users wouldn't have to quote their space separated paths.

Another approach could be to keep the `path` parameter and introduce a `dependencies` optional variadic parameter.

test/general-tests/starknet-verify/check.sh
```diff
@@ -4,13 +4,24 @@ set -e
 echo "The starknet-verify test is temporarily disabled."
 echo "To enable it back, uncomment the lines in its check.sh."
 
-# npx hardhat starknet-compile contracts/util.cairo
+
+# npx hardhat starknet-compile contracts/contract.cairo contracts/util.cairo
```
FabijanC, line 8: It should be commented here why the util contract is compiled as well.

test/general-tests/starknet-verify/check.sh
```diff
@@ -4,13 +4,24 @@ set -e
...
 # output=$(npx hardhat starknet-deploy --starknet-network "$NETWORK" starknet-artifacts/contracts/util.cairo/)
-# echo $output
-# address=$(echo $output | sed -r "s/.*Contract address: (\w*).*/\1/")
+# util_address=$(echo $output | sed -r "s/.*Contract address: (\w*).*/\1/")
+
+# output=$(npx hardhat starknet-deploy --starknet-network "$NETWORK" starknet-artifacts/contracts/contract.cairo/)
+# main_address=$(echo $output | sed -r "s/.*Contract address: (\w*).*/\1/")
 
-# echo "Address: $address"
+# echo "Util contract address: $util_address"
```
FabijanC, line 16: You can `echo` this right after the deployment.

src/task-actions.ts
```diff
@@ -291,34 +297,100 @@ export async function starknetVoyagerAction(args: TaskArguments, hre: HardhatRun
...
+    paths.forEach(function (item, index) {
+        if (!path.isAbsolute(item)) {
+            paths[index] = path.normalize(path.join(hre.config.paths.root, item));
+            if (!fs.existsSync(paths[index])) {
+                throw new HardhatPluginError(PLUGIN_NAME, `File ${paths[index]} does not exist`);
+            }
+        }
+    });
+
+    const bodyFormData = new URLSearchParams();
```
FabijanC, line 323: I find it unintuitive that for `bodyFormData` is constructed using `URLSearchParams`. Is this really the way?

src/task-actions.ts
```diff
@@ -291,34 +297,100 @@ export async function starknetVoyagerAction(args: TaskArguments, hre: HardhatRun
...
+                throw new HardhatPluginError(PLUGIN_NAME, msg);
+            }
+            default: {
+                const msg = `Something went wrong when trying to verify the code at address ${address}`;
+                throw new HardhatPluginError(PLUGIN_NAME, msg);
+            }
+        }
+    });
+    console.log(`Contract has been successfuly verified at address ${address}`);
+    return;
```
FabijanC, line 358: why return?

src/task-actions.ts
```diff
@@ -291,34 +297,100 @@ export async function starknetVoyagerAction(args: TaskArguments, hre: HardhatRun
...
-    let contractPath = args.path;
+    if (args.path) {
+        await handleSingleContractVerification(args.path, args.address, voyagerUrl, hre);
+    } else {
+        const splitPaths = args.paths.split(" ");
+        await handleMultiPartContractVerification(splitPaths, args.address, voyagerUrl, hre);
+    }
+}
+
+async function handleMultiPartContractVerification(
```
FabijanC, line 308: These two functions should be refactored. There is a lot of common functionality like path checking and response checking.

FabijanC (changes requested):

src/task-actions.ts
dribeiro-ShardLabs: the other option for the formData would be to add a new dependency 'form-data', but the URLSearchParams works exactly the same

src/index.ts
```diff
@@ -203,6 +203,12 @@ task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
     .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
     .addParam("address", "The address where the contract is deployed")
+    .addOptionalVariadicPositionalParam(
+    "paths",
+    "The paths of the dependencies of the contract specified in --param" +
```
FabijanC, line 208: --param? or --path?

src/index.ts
```diff
@@ -203,6 +203,12 @@ task("starknet-verify", "Verifies the contract in the Starknet network.")
     .addOptionalParam("starknetNetwork", "The network version to be used (e.g. alpha)")
     .addParam("path", "The path of the cairo contract (e.g. contracts/conract.cairo)")
```
FabijanC, line 204: Would adding the adjective "main" to "contract" be useful?

FabijanC (changes requested):

FabijanC (approved):

PR #72, opened 2022-03-07
Automatic account artifacts fetching

dribeiro-ShardLabs:
Fetching the artifacts for the account contract automatically, so the contract name parameter is removed from the functions `deployFromABI` and `getAccountFromAddress`

### DIFF
--- a/src/account-utils.ts
+++ b/src/account-utils.ts
@@ -1,0 +1,124 @@
+import { Numeric } from "./types";
+import { hash } from "starknet";
+import { BigNumberish, toBN } from "starknet/utils/number";
+import * as ellipticCurve from "starknet/utils/ellipticCurve";
+import { ec } from "elliptic";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import * as fs from "fs";
+import path from "path";
+import {
+    ABI_SUFFIX,
+    ACCOUNT_ARTIFACTS_VERSION,
+    ACCOUNT_CONTRACT_ARTIFACTS_ROOT_PATH,
+    GITHUB_ACCOUNT_SHA_URL
+} from "./constants";
+import axios from "axios";
+
+/*
+ * Helper cryptography functions for Key generation and message signing
+ */
+
+export function generateRandomStarkPrivateKey(length = 63) {
+    const characters = "0123456789ABCDEF";
+    let result = "";
+    for (let i = 0; i < length; ++i) {
+        result += characters.charAt(Math.floor(Math.random() * characters.length));
+    }
+    return toBN(result, "hex");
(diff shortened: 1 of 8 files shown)


src/account-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+        targetPath
+    );
+
+    if (!fs.existsSync(artifactsTargetPath)) {
+        const baseArtifactsPath = path.join(
+            hre.config.paths.starknetArtifacts,
+            ACCOUNT_CONTRACT_ARTIFACTS_ROOT_PATH
+        );
+        if (fs.existsSync(baseArtifactsPath)) {
+            fs.rmSync(baseArtifactsPath, { recursive: true, force: true });
```
FabijanC, line 87: There could be a test asserting the unneeded files/directories are indeed deleted.

src/account-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+            abi: artifactsName + ABI_SUFFIX
+        };
+
+        fs.mkdirSync(artifactsTargetPath, { recursive: true });
+
+        const githubTree = GITHUB_ACCOUNT_SHA_URL.concat("/", accountType, "/", targetPath);
+
+        let jsonBlobUrl, abiBlobUrl;
+
+        // Get the url to the file blobs through the github api tree, because if the files are too large, the regular method of retrieving will not work
```
FabijanC, line 101: if raw.usergithubcontent.com can indeed be used, the logic for a single file can be extracted to a dedicated function.

src/account-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+        await axios({
```
FabijanC, line 102: Why await and .then?

FabijanC (changes requested):

src/account-utils.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (fs.existsSync(baseArtifactsPath)) {
+            fs.rmSync(baseArtifactsPath, { recursive: true, force: true });
+        }
+
+        const jsonArtifact = artifactsName + ".json";
+        const abiArtifact = artifactsName + ABI_SUFFIX;
+
+        fs.mkdirSync(artifactsTargetPath, { recursive: true });
+
+        const fileLocationUrl = GITHUB_ACCOUNT_ARTIFACTS_URL.concat("/", accountType, "/", targetPath, "/");
```
FabijanC, line 95: Prepending a slash can be avoided by having it already in the constant.

FabijanC (changes requested):

FabijanC (approved):

PR #78, opened 2022-03-10
Hotfix for account artifacts, return tx hash on invoke

dribeiro-ShardLabs:
## Usage related changes
- Users no longer need to explicitly create a `starknet-artifacts/account-contract-artifacts` folder to avoid the error `Error: ENOENT: no such file or directory`

- Check if the account artifacts exist before accessing the folder;
- Account invoke() return is of type `type InvokeResponse = string`, and it returns the transaction hash of the invoke.

# Checklist:

### DIFF
--- a/src/account-utils.ts
+++ b/src/account-utils.ts
@@ -68,66 +68,67 @@ export async function handleAccountContractArtifacts(
     hre: HardhatRuntimeEnvironment
 ): Promise<string> {
     // Name of the artifacts' parent folder
-    const targetPath = artifactsName + ".cairo";
+    const artifactsBase = artifactsName + ".cairo";
+
+    const baseArtifactsPath = path.join(
+        hre.config.paths.starknetArtifacts,
+        ACCOUNT_CONTRACT_ARTIFACTS_ROOT_PATH
+    );
+
+    // Remove old versions from the path
+    if (fs.existsSync(baseArtifactsPath)) {
+        const contents = fs.readdirSync(baseArtifactsPath);
+        contents
+            .filter((content) => content !== ACCOUNT_ARTIFACTS_VERSION)
+            .forEach((content) => {
+                fs.rmSync(path.join(baseArtifactsPath, content), {
+                    recursive: true,
+                    force: true
+                });
+            });
+    }
 
     // Full path to where the artifacts will be saved
     const artifactsTargetPath = path.join(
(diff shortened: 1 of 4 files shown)


src/account-utils.ts
```diff
@@ -66,62 +66,69 @@ export async function handleAccountContractArtifacts(
...
-        transformResponse: (res) => {
-            return res;
-        },
-        responseType: "json"
-    });
-
-    fs.writeFileSync(path.join(artifactsTargetPath, artifactName), response.data);
+    if (!fs.existsSync(path.join(artifactsTargetPath, artifact))) {
+        // Check if an old version of the artifacts still exists in the path, if so delete it
+        fs.mkdirSync(artifactsTargetPath, { recursive: true });
```
FabijanC, line 121: Will this create problems if the path already exists?

src/account-utils.ts
```diff
@@ -66,62 +66,69 @@ export async function handleAccountContractArtifacts(
...
+        artifactsBase
+    );
 
-        const jsonArtifact = artifactsName + ".json";
-        const abiArtifact = artifactsName + ABI_SUFFIX;
+    const jsonArtifact = artifactsName + ".json";
+    const abiArtifact = artifactsName + ABI_SUFFIX;
 
-        fs.mkdirSync(artifactsTargetPath, { recursive: true });
+    const fileLocationUrl = GITHUB_ACCOUNT_ARTIFACTS_URL.concat(
```
FabijanC, line 101: The variable name suggests this is a URL to a file, whereas it's a directory.

src/account-utils.ts
```diff
@@ -66,62 +66,69 @@ export async function handleAccountContractArtifacts(
...
 
-        await downloadArtifact(jsonArtifact, artifactsTargetPath, fileLocationUrl);
-        await downloadArtifact(abiArtifact, artifactsTargetPath, fileLocationUrl);
-    }
+    return artifactsTargetPath;
 }
-async function downloadArtifact(
-    artifactName: string,
+
+async function assertArtifact(
```
FabijanC, line 114: Add JSDoc to explain what this method does.

FabijanC (changes requested):
How does the plugin behave if there is no starknet-artifacts directory?

FabijanC (approved):

PR #84, opened 2022-03-28
Invoke options

FabijanC:
## Usage related changes
- Allow specifying maxFee and nonce.
- Introduce `--account-contract` flag for compiling account contracts.
  - Account contracts are currently those with a function called `__execute__`, hence the addition of such a test. Couldn't make it 

- Enable wallet testing on devnet.
- Revert to es2016 in tsconfig.json (es2020 was never a part of an official plugin release, but happened in a commit between the latest release and this commit).
- Introduce `copyWithBigint` which basically takes an already existing functionality and wraps it into a function. The function is now used in one more place.
- Make Account methods protected.
- Unfortunately some hackiness is introduced in multiInteract in account.ts, if the reviewer should find a better way of dealing with it
- Refactor InteractChoice making it a class with methods and properties - pushes some logic directly into InteractChoice instead of having to write ifs and elses in various places.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ If no paths are provided, all Starknet contracts in the default contracts direct
 
 `--cairo-path` allows specifying the locations of imported files, if necessary. Separate them with a colon (:), e.g. `--cairo-path='path/to/lib1:path/to/lib2'`
 
-`--account-contract` allows compiling a contract containing account implementation.
+`--account-contract` allows compiling an account contract.
 
 ### `starknet-deploy`
 
--- a/README.md
+++ b/README.md
@@ -48,12 +48,14 @@ This plugin defines the following Hardhat commands (also called tasks):
 ### `starknet-compile`
 
 ```
-npx hardhat starknet-compile [PATH...] [--cairo-path "<LIB_PATH1>:<LIB_PATH2>:..."]
+npx hardhat starknet-compile [PATH...] [--cairo-path "<LIB_PATH1>:<LIB_PATH2>:..."] [--account-contract]
 ```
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -48,13 +48,15 @@ This plugin defines the following Hardhat commands (also called tasks):
...
 ```
-npx hardhat starknet-compile [PATH...] [--cairo-path "<LIB_PATH1>:<LIB_PATH2>:..."]
+npx hardhat starknet-compile [PATH...] [--cairo-path "<LIB_PATH1>:<LIB_PATH2>:..."] [--account-contract]
 ```
 
 If no paths are provided, all Starknet contracts in the default contracts directory are compiled. Paths can be files and directories.
 
 `--cairo-path` allows specifying the locations of imported files, if necessary. Separate them with a colon (:), e.g. `--cairo-path='path/to/lib1:path/to/lib2'`
 
+`--account-contract` allows compiling a contract containing account implementation.
```
dribeiro-ShardLabs, line 58: I think it should be specified that it is a flag that when set, it indicates that the contract provided is an account contract

src/index.ts
```diff
@@ -148,6 +153,7 @@ task("starknet-compile", "Compiles Starknet contracts")
         "Allows specifying the locations of imported files, if necessary.\n" +
             "Separate them with a colon (:), e.g. --cairo-path='path/to/lib1:path/to/lib2'"
     )
+    .addFlag("accountContract", "Use this flag if you are compiling an account contract")
```
dribeiro-ShardLabs, line 156: Same as described in the readme comment

PR #95, opened 2022-04-20
Fix Devnet caching

FabijanC:
## Usage related changes
None

- So far devnet installation and caching were not synchronised (installation was using some fixed version from pypi, caching was checking the latest github commit of the devnet repo). This is now fixed.

# Checklist:

### DIFF
--- a/scripts/get-version.py
+++ b/scripts/get-version.py
@@ -1,4 +1,4 @@
-#!/bin/python3
+#!/usr/bin/python3
 
 """
 Returns the version used in this project of the tool provided as a CLI argument.
--- a/scripts/get-version.py
+++ b/scripts/get-version.py
@@ -1,0 +1,22 @@
+#!/bin/python3
+
+"""
+Returns the version used in this project of the tool provided as a CLI argument.
+"""
+
+import sys
+
+if len(sys.argv) != 2:
+    sys.exit(sys.argv[0] + " <TOOL>")
+
+tool = sys.argv[1].strip()
+
+TOOLS_USED = {
+    "starknet-devnet": "0.1.22"
+}
+
+if tool not in TOOLS_USED:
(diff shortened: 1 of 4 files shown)


scripts/install-devnet.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-pip3 install starknet-devnet==0.1.22
+pip3 install starknet-devnet==$(../scripts/get-version.py starknet-devnet)
```
badurinantun, line 5: It would cleaner if version was in a variable

scripts/get-version.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+import sys
+
+if len(sys.argv) != 2:
+    sys.exit(sys.argv[0] + " <TOOL>")
+
+tool = sys.argv[1].strip()
+
+TOOLS_USED = {
```
badurinantun, line 14: This could also be a JSON file, then we don't need have a script but can use `jq`.

Also, perhaps if we had it in the some JSON file we could use that for the constants as well (where we specify the cairo version, but will also specify a devnet version in the future)

badurinantun (changes requested):

scripts/get-version.py
FabijanC: At first it was a JSON parsed with `jq`, but then I realized that way we would need to have the `jq` logic in multiple places.

FabijanC: As agreed, this can stay as it is and will be handled in a different PR.

scripts/install-devnet.sh
FabijanC: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/95/commits/3b668f233ce0ad839e9f0994f922f53d497ba3e6

PR #108, opened 2022-05-13
Adapt Account to OpenZeppelin 0.1.0 and Argent 0.2.1

FabijanC:
## Usage related changes
- Adapt to [OpenZeppelin account version 0.1.0](https://github.com/OpenZeppelin/cairo-contracts/tree/v0.1.0/src/openzeppelin/account) and [Argent account version 0.2.1](https://github.com/argentlabs/argent-contracts-starknet/tree/v0.2.1/contracts).
- Minor updates to existing Readme parts.
- Document used account versions in Readme

- Read account artifacts from the plugin repo (not from the example repo)
- Introduced rawOutput manipulation because the Argent implementation returns raw output from its `__execute__`
- Rename handleMultiCall -> handleMultiInteract and make it a common Account method.

# Checklist:

- [x] I **didn't** have to update the tests
- [x] I **didn't** have to create a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/Shard-Labs/starknet-hardhat-example).


src/account-utils.ts
```diff
@@ -132,34 +63,21 @@ export async function handleAccountContractArtifacts(
...
-        ACCOUNT_ARTIFACTS_VERSION,
+        accountType,
+        artifactsVersion,
         artifactsBase
     );
 
     const jsonArtifact = artifactsName + ".json";
     const abiArtifact = artifactsName + ABI_SUFFIX;
 
     const artifactLocationUrl = GITHUB_ACCOUNT_ARTIFACTS_URL.concat(
```
badurinantun, line 77: Why are we using Github repo for this? Why don't we just load the artifacts from the directory?

FabijanC: Changed downloading from github to copying from the plugin: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/108/commits/188a6e27ac2a391812d3854ccbc6b28e780d2e3d
Set to unminifed artifacts with debug info: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/108/commits/c78632c4c863a14df71c73f35129131a6a46b64a

src/account.ts
```diff
@@ -388,15 +547,19 @@ export class ArgentAccount extends Account {
...
         return "__execute__";
     }
 
-    async getNonce(): Promise<bigint> {
+    protected async getNonce(): Promise<bigint> {
         const { nonce } = await this.starknetContract.call("get_nonce");
         return nonce;
     }
+
+    protected hasRawOutput(): boolean {
```
badurinantun, line 562: Why is this a function?

src/account-utils.ts
```diff
@@ -58,142 +44,61 @@ export function signMultiCall(
...
-                fs.rmSync(path.join(baseArtifactsPath, content), {
-                    recursive: true,
-                    force: true
-                });
-            });
-    }
+    const baseArtifactsPath = path.join(hre.config.paths.starknetArtifacts, ACCOUNT_ARTIFACTS_DIR);
 
     // Full path to where the artifacts will be saved
     const artifactsTargetPath = path.join(
```
badurinantun, line 59: Why are artifacs being saved to a target directory?

src/account.ts
FabijanC: If you're asking why it isn't a property, it doesn't have to be, it's just following the same paradigm as `getExecutionFunctionName`, which could also be a property set through constructor.

src/account-utils.ts
FabijanC: Accounts are also contracts. So for the plugin to be able to interact with them as contracts, their artifacts need to be in the artifacts directory. Internally, doing `account.call(...)` does something like `account.starknetContract.call(...)`

PR #111, opened 2022-05-17
Cairo lang 0.8.2

FabijanC:
## Usage related changes
- Adapt to cairo-lang 0.8.2
- Default to the latest block in getBlock
- Update to devnet 0.2.1

- Introduce type BlockNumber that can be `number`, `"pending"`, or `"latest"`

TODO:
- Should defaulting to `pending` be replaced with defaulting to `latest`?
- Where else is defaulting not implemented?

# Checklist:

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
     "CAIRO_LANG": "0.8.2",
-    "STARKNET_DEVNET": "0.2.0"
+    "STARKNET_DEVNET": "0.2.1"
 }
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "CAIRO_LANG": "0.8.1",
+    "CAIRO_LANG": "0.8.2",
     "STARKNET_DEVNET": "0.2.0"
 }
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "CAIRO_LANG": "0.8.2",
+    "CAIRO_LANG": "0.8.2.1",
     "STARKNET_DEVNET": "0.2.1"
 }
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -598,6 +599,7 @@ export class StarknetContract {
         args?: StringMap,
         options: EstimateFeeOptions = {}
     ): Promise<FeeEstimation> {
+        options = copyWithBigint(options); // copy because of potential changes to the object
```
badurinantun, line 602: If have to add comments that means this code is hard to understand by itself. We shouldn't change the function arguments, but create a new object with the correct block number

config.json
```diff
@@ -1,4 +1,4 @@
 {
-    "CAIRO_LANG": "0.8.1",
-    "STARKNET_DEVNET": "0.2.0"
+    "CAIRO_LANG": "0.8.2",
```
badurinantun, line 2: What about `0.8.2.1`?

badurinantun (changes requested):

config.json
FabijanC: totally forgot about that

src/types.ts
FabijanC: this is actually old code, just noticed that it was missing the copying part; but I can try to edit it

config.json
FabijanC: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/111/commits/e9cbce03eb3528eee7c8d7d7ed6ae01a40385215

src/types.ts
FabijanC: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/111/commits/4b20ead70affda21c4b2ba9f2ed6296ea33bdc01

badurinantun (approved):

PR #114, opened 2022-05-30
Time manipulation functions

Nathan-SL:
## Usage related changes
- Block timestamp can be handled using the two devnet functions `increaseTime()` and `setTime()`

- `setTime()` wil update the subsequent block timestamp keeping the offset

# Checklist:

### DIFF
--- a/src/devnet-utils.ts
+++ b/src/devnet-utils.ts
@@ -35,8 +35,16 @@ export interface LoadL1MessagingContractResponse {
     l1_provider: string;
 }
 
+export interface SetTimeResponse {
+    next_block_timestamp: number;
+}
+
+export interface IncreaseTimeResponse {
+    timestamp_increased_by: number;
+}
+
 export class DevnetUtils implements Devnet {
-    constructor(private hre: HardhatRuntimeEnvironment) {}
+    constructor(private hre: HardhatRuntimeEnvironment) { }
 
     private get endpoint() {
         return `${this.hre.config.starknet.networkUrl}`;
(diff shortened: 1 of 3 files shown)


src/devnet-utils.ts
```diff
@@ -79,4 +87,26 @@ export class DevnetUtils implements Devnet {
...
+            const response = await axios.post<IncreaseTimeResponse>(
+                `${this.endpoint}/increase_time`,
+                {
+                    time: seconds
+                });
+            return response.data;
+        }, "Request failed. Make sure your network has the /increase_time endpoint");
+    }
+
+    public async setTime(seconds: number) {
```
badurinantun, line 102: Do you think it would make sense to provide an option to set time in `Date` as well? Or does it make more sense to expect users to do the date manipulation anyway?

badurinantun (changes requested):
Can you add tests for this as well?

src/type-extensions.ts
```diff
@@ -121,6 +121,21 @@ declare module "hardhat/types/runtime" {
...
+
+        /**
+         * Increases the block timestamp offset.
+         * @param seconds the new timestamp in seconds
+         * @returns a timestamp
+         */
+        increaseTime: (seconds: number) => Promise<IncreaseTimeResponse>;
+
+        /**
+         * Sets the block timestamp offset.
```
FabijanC, line 134: This JSDoc is wrong. It is not the offset which is set, but the timestamp itself.

src/type-extensions.ts
```diff
@@ -121,6 +121,21 @@ declare module "hardhat/types/runtime" {
             address?: string,
             networkId?: string
         ) => Promise<LoadL1MessagingContractResponse>;
+
+
+        /**
+         * Increases the block timestamp offset.
+         * @param seconds the new timestamp in seconds
```
FabijanC, line 128: This JSDoc is wrong. It is not the timestamp which is is set, but the offset which is increased.

FabijanC (changes requested):

PR #123, opened 2022-06-17
Arm docker image

Nathan-SL:
## Usage related changes
-  For `arm` machines `-arm` image will be pulled. On images `integrated-devnet` and `cairo-cli`.

- 

## Checklist:

- [ ] I have updated the tests
- [ ] I have created a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/Shard-Labs/starknet-hardhat-example).

### DIFF
--- a/src/devnet/create-devnet-wrapper.ts
+++ b/src/devnet/create-devnet-wrapper.ts
@@ -8,7 +8,7 @@ import {
     INTEGRATED_DEVNET_URL,
     PLUGIN_NAME
 } from "../constants";
-import { getNetwork } from "../utils";
+import { getImageTagByArch, getNetwork } from "../utils";
 import { DockerDevnet } from "./docker-devnet";
 import { VenvDevnet } from "./venv-devnet";
 import { IntegratedDevnet } from "./integrated-devnet";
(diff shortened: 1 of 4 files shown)


src/devnet/create-devnet-wrapper.ts
```diff
@@ -40,10 +40,17 @@ export function createIntegratedDevnet(hre: HardhatRuntimeEnvironment): Integrat
         );
     }
 
+    // Check CPU architecture
+    const arch = process.arch;
+    let tag = devnetNetwork.dockerizedVersion || DEFAULT_DEVNET_DOCKER_IMAGE_TAG;
+    if (arch.includes("arm")) {
```
FabijanC, line 46: 1) Our images are arm64 compatible. I don't know if that also means they are arm compatible - investigate this. Also, instead of using `arch.includes`, rather use something like `["arm", "arm64"].includes` (although, depending on your findings, this might change to just checking if `arch === "arm64"`).
2) Extract the arch checking and appending of the suffix to a util function.
3) Don't use `tag.includes` but `tag.endsWith`.

FabijanC (changes requested):

src/devnet/create-devnet-wrapper.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/123/commits/3cef8f790f91f05e5332c46e3f7cced0ec208fe0

src/utils.ts
```diff
@@ -225,3 +225,12 @@ export function copyWithBigint<T>(object: unknown): T {
         )
     );
 }
+
+export function getImageTagByArch(tag: string): string {
+    // Check CPU architecture
+    const arch = process.arch;
+    if (arch === "arm64") {
+        tag = tag.endsWith("-arm") ? tag : `${tag}-arm`;
```
FabijanC, line 233: Instead of using the ternary operator, it could be checked if arch is arm64 and if tag doesn't end with -arm in the same if-clause.

FabijanC (changes requested):

src/utils.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/123/commits/be0ea87ac6db94e860ea92f56e799b16af6562dc

FabijanC (approved):

FabijanC:
Skipping ci on merge commit because this is only relevant for integrated-devnet, which already passed ci tests.

PR #132, opened 2022-06-30
Fix exiting in integrated-devnet

FabijanC:
## Usage related changes

-   Log message if integrated-devnet exits with status code not equal to zero.

-   Added tests that expect integrated-devnet failure (if wrong cli parameters or address occupied, for both venv and docker).
-   Removed the `--detach` flag from the dockerized integrated devnet
  -   It is still handled correctly, but allows us to use the subprocess return by the docker (with --detach, it would spawn the container in background and that spawner process would die immediatelly after spawning)
-   Drew inspiration from #133.

## Checklist:

-   [x] Formatted the code
-   [x] No linter errors
-   [x] Updated the `test` directory
-   [x] All tests are passing


Nathan-SL (approved):
Great work!

PR #138, opened 2022-07-07
Fix logging

### LINKED ISSUE #136: Remove double logging
Somewhere we are probably doing log and throw, so errors get printed twice. If this is the case, only throw, don't also log. If this is not the case and we cannot do anything about it, report here.

FabijanC:
## Usage related changes

- Closes #136:
  - This stems from multiple sources, just one is log-and-throw done in the plugin, the rest comes from the way Starknet CLI handles errors (this was addressed in a hacky way, by pattern matching) 
- Improve error logging if wrong type provided as function input
- Update docs with `getPredeployedAccounts` usage.
- Improve formatting of error messages (unescape escaped newlines)

- Refactor event decoding:
  - incorporate `adaptEvent` into `decodeEvents`
  - introduce type `EventAbi`

## Checklist:

-   [x] Formatted the code
-   [x] No linter errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,12 +637,14 @@ await account.invoke(contract, "increase_balance", { amount });
     -   Give it finds through [the faucet](https://faucet.goerli.starknet.io/).
     -   Later load the account using `starknet.getAccountFromAddress`.
 -   **On starknet-devnet**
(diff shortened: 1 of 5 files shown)


Nathan-SL (approved):
LGTM!

PR #172, opened 2022-07-20
Add new param to voyager verifier API

Nathan-SL:
## Usage related changes

-   The Voyager verification API expects `account-contract` as a param.

-   NA

## Checklist:

-   [x] Formatted the code
-   [x] No linter errors
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -267,6 +267,7 @@ task("starknet-verify", "Verifies a contract on a Starknet network.")
     .addParam("path", "The path of the main cairo contract (e.g. contracts/contract.cairo)")
     .addParam("address", "The address where the contract is deployed")
     .addParam("compilerVersion", "The compiler version used to compile the cairo contract")
+    .addOptionalParam("accountContract", "The contract type which specifies whether it's an account contract. Omiting it sets false.")
     .addOptionalParam("license", "The licence of the contract (e.g No License (None))")
     .addOptionalVariadicPositionalParam(
         "paths",
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -338,6 +338,7 @@ async function handleContractVerification(
 
     const bodyFormData = new FormData();
     bodyFormData.append("compiler-version", args.compilerVersion);
+    bodyFormData.append("account-contract", args.accountContract ? "true" : "false");
     bodyFormData.append("license", args.license || "No License (None)");
 
     // Dependencies (non-main contracts) are in args.paths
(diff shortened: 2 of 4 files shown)


src/index.ts
```diff
@@ -267,6 +267,7 @@ task("starknet-verify", "Verifies a contract on a Starknet network.")
     .addParam("path", "The path of the main cairo contract (e.g. contracts/contract.cairo)")
     .addParam("address", "The address where the contract is deployed")
     .addParam("compilerVersion", "The compiler version used to compile the cairo contract")
+    .addOptionalParam("accountContract", "The contract type which specifies whether it's an account contract. Omiting it sets false.")
```
FabijanC, line 270: Do we need to specify that it's a boolean param somewhere?

src/task-actions.ts
```diff
@@ -338,6 +338,7 @@ async function handleContractVerification(
 
     const bodyFormData = new FormData();
     bodyFormData.append("compiler-version", args.compilerVersion);
+    bodyFormData.append("account-contract", args.accountContract ? "true" : "false");
```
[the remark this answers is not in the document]
FabijanC: Also, what is the value of `args.accountContract` if the user specifies it, and what's the value if it's not specified?

src/index.ts
FabijanC: Also, there's a typo, should be `omitting`

src/task-actions.ts
```diff
@@ -338,6 +338,13 @@ async function handleContractVerification(
 
     const bodyFormData = new FormData();
     bodyFormData.append("compiler-version", args.compilerVersion);
+    let accountContract;
+    if (args.accountContract && args.accountContract === true) {
```
FabijanC, line 342: This doesn't look good:
- it's redundant to check both `args.accountContract` and `args.accountContract === true`, the latter should be enough
- this condition will never be met because the user cannot specify a boolean value through the CLI, right?
- this could as well be achieved with a ternary operator
- this if-else structure means that anything other than `true` will be treated as false, do we want that?

FabijanC (changes requested):

src/task-actions.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/172/commits/c8ecc43a6a76744cc346bba1728c66a49caa916a

FabijanC (approved):

PR #175, opened 2022-07-22
Use proxy server in VenvWrapper

FabijanC:
## Usage related changes

- Should only present a speed-up; no change in functionality.

- Spawn a server that acts as an intermediary between user requests and Starknet CLI
  - `VenvWrapper` is modified to not spawn subprocesses (for most commands)
- Create a new abstract class `ExternalServer` which abstracts the previous `IntegratedDevnet` and the new `StarknetVenvProxy`.
- Something similar can be done for `DockerWrapper`. but in a separate PR (might include changing the cairo-cli docker image so that it includes the proxy server script).
- Not using `"python"` command directly in `VenvWrapper`, it is adapted in the constructor according to what the user specified as `venv` (`active` or not)
  - Now using proprety `pythonPath`

- Currently available on npm (`@sharlabs/starknet-hardhat-plugin`) and in [tags](https://github.com/Shard-Labs/starknet-hardhat-plugin/tags) as alpha versions:
  - `0.6.3-alpha` - CLI proxy listens on port 8080 (hardcoded)
  - `0.6.3-alpha.1` - CLI proxy listens on a randomly selected free port (subject to race condition)
  - `0.6.3-alpha.2` - CLI proxy listens on a port determined by polling ports 6050 (devnet default +1000), 7050, 8050, ... (stops on first free port)
    - what could be tried out (but would require more refactoring) is taking into accout if devnet is being used and use its port as the starting point for port search

## Checklist:

-   [x] Formatted the code
-   [x] No linter errors
-   [x] Restored version in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
         "test-general-tests": "TEST_SUBDIR=general-tests ./scripts/test.sh",
         "test-venv-tests": "TEST_SUBDIR=venv-tests ./scripts/test.sh",
         "test-integrated-devnet-tests": "TEST_SUBDIR=integrated-devnet-tests ./scripts/test.sh",
-        "build": "rm -rf dist && tsc",
+        "build": "rm -rf dist && tsc && npm run copy-files",
+        "copy-files": "cp src/starknet_cli_wrapper.py dist/src/",
         "lint": "eslint src test"
     },
     "author": "Shard-Labs",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@shardlabs/starknet-hardhat-plugin",
-    "version": "0.6.2",
+    "version": "0.6.3-alpha",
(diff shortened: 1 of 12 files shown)


Nathan-SL (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@shardlabs/starknet-hardhat-plugin",
-    "version": "0.6.2",
+    "version": "0.6.3-alpha.2",
```
ivpavici, line 3: @Nathan-SL @FabijanC you agreed on using this tag out of the 3 mentioned options? I'm asking because the PR is approved!

FabijanC: The version that is currently on branch `proxy-script` (the branch to be merged with this PR) is the one that was released as `0.6.3-alpha.2`

PR #181, opened 2022-08-05
Starknet plugin version

### LINKED ISSUE #163: Add hardhat task that returns currently installed plugin version
Something that would be used like this:

```
npx hardhat starknet-plugin-version
```

and return

```
0.6.1
```

or whichever is the current version.

This would be useful when debugging with users (so that they don't have to interpret their package.json, package-lock.json or yarn.lock files when asked about the plugin version they are using).

Nathan-SL:
## Usage related changes

-  Plugin version can be printed using task `npx hardhat starknet-plugin-version`
-  Closes #163 

-   NA

## Checklist:

-   [x] Formatted the code
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -595,7 +595,6 @@ export async function starknetRunAction(
     });
 }
 
-export async function starknetPluginVersionAction(
-) {
+export async function starknetPluginVersionAction() {
     console.log(`Version: ${version}`);
 }
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -584,3 +584,11 @@ export async function starknetRunAction(
         await runSuper(args);
     });
 }
+
+export async function starknetPluginVersionAction(
+) {
+    const version = JSON.parse(
+        fs.readFileSync(path.join(__dirname, "../../", "package.json"), "utf8")
+    ).version;
+    console.log(`Version: ${version}`);
+}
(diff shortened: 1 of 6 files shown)


src/task-actions.ts
```diff
@@ -594,3 +594,11 @@ export async function starknetRunAction(
         await runSuper(args);
     });
 }
+
+export async function starknetPluginVersionAction(
+) {
+    const version = JSON.parse(
```
FabijanC, line 600: Can you consider something like this: https://stackoverflow.com/questions/9153571/is-there-a-way-to-get-version-from-package-json-in-nodejs-code

test/general-tests/starknet-plugin-version/check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+# Check if plugin version returns the correct version
+PLUGIN_VERSION=$(npm view @shardlabs/starknet-hardhat-plugin dist-tags.latest)
```
FabijanC, line 5: Imagine if a commit is made where the version in package.json is incremented. Wouldn't then this test fail, even though it shouldn't?

FabijanC (changes requested):

src/task-actions.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/181/commits/30653a3ad9cfab02eaf9139abe1df9471489fcc7

test/general-tests/starknet-plugin-version/check.sh
Nathan-SL: That's right. I think running the command would suffice. https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/181/commits/19453a442c1abff42ae2b47cfca7f103d04df6c3

PR #191, opened 2022-08-19
Use dockerized plugin through proxy server

### LINKED ISSUE #184: Use a proxy server for dockerized plugin
Similar to how using a proxy server for the venv version of plugin was implemented in https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/175.

Ideally there would be a proxy server started on hardhat startup and killed on hardhat shutdown. The server would have starknet CLI preloaded and it would accept requests with arguments for `starknet` or `starknet-compile`. This is in contrast to the current approach where a new `docker run` is executed each time when a starknet CLI command is needed.

Some commands which are not expected to be called intensively during a test may be executed directly on the host machine (like starknet-compile), but there could be a way to handle them too.

If path binding from host machine to the docker container is a problem, consider either copying the files in the container (through the server) or sending them together with the rest of the payload.

FabijanC:
## Usage related changes

- Speed up dockerized plugin
- Close #184
- Remove docker server logging on startup and shutdown

- Rely on a proxy server in `DockerWrapper`
- Refactor external server and its derived classes
- Append random suffix to docker container name (to allow multiple containers run at the same time)
- Speed up tests (by about 25%)

## Checklist:

-   [x] Formatted the code


src/external-server/docker-devnet.ts
```diff
@@ -0,0 +1,23 @@
...
+        host: string,
+        port: string,
+        private devnetArgs?: string[],
+        stdout?: string,
+        stderr?: string
+    ) {
+        super(image, host, port, "is_alive", "integrated-devnet", devnetArgs, stdout, stderr);
+    }
+
+    protected async getDockerArgs(): Promise<string[]> {
```
Nathan-SL, line 16: `getDockerArgs()` don't need async modifier.

src/starknet-docker-proxy.ts
```diff
@@ -0,0 +1,42 @@
...
+    constructor(
+        image: Image,
+        private rootPath: string,
+        private accountPaths: string[],
+        private cairoPaths: string[]
+    ) {
+        super(image, "127.0.0.1", null, "", "starknet-docker-proxy");
+    }
+
+    protected async getDockerArgs(): Promise<string[]> {
```
Nathan-SL, line 26: Here, no need for the async modifier.

FabijanC: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/191/commits/ce20a6f341cd4d34620c42c9603d4753336bae8b

src/external-server/docker-devnet.ts
FabijanC: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/191/commits/ce20a6f341cd4d34620c42c9603d4753336bae8b

Nathan-SL (approved):

PR #202, opened 2022-08-29
Random port on integrated devnet

### LINKED ISSUE #176: Use random free port for integrated-devnet
With the `getFreePort` function introduced in https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/175

Nathan-SL:
## Usage related changes

-   Uses free random ports on integrated devnet
-   Closes #176 

-   NA

## Checklist:

-   [x] Formatted the code
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/task-actions.ts
+++ b/src/task-actions.ts
@@ -553,7 +553,7 @@ async function runWithDevnet(hre: HardhatRuntimeEnvironment, fn: () => Promise<u
         return;
     }
 
-    const devnet = createIntegratedDevnet(hre);
+    const devnet = await createIntegratedDevnet(hre);
 
     await devnet.start();
     await fn();
--- a/src/external-server/create-devnet-wrapper.ts
+++ b/src/external-server/create-devnet-wrapper.ts
@@ -10,15 +10,16 @@ import {
 import { getImageTagByArch, getNetwork } from "../utils";
 import { DockerDevnet } from "./docker-devnet";
 import { VenvDevnet } from "./venv-devnet";
-import { ExternalServer } from "./external-server";
+import { ExternalServer, getFreePort } from "./external-server";
 
-export function createIntegratedDevnet(hre: HardhatRuntimeEnvironment): ExternalServer {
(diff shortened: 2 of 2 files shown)


src/external-server/docker-devnet.ts
```diff
@@ -18,6 +18,7 @@ export class DockerDevnet extends DockerServer {
     }
 
     protected async getContainerArgs(): Promise<string[]> {
-        return this.devnetArgs || [];
+        const containerArgs = this.devnetArgs || [];
```
FabijanC, line 21: What happens if the user provided `--port` in `args` in hardhat config?

src/task-actions.ts
```diff
@@ -554,9 +554,14 @@ async function runWithDevnet(hre: HardhatRuntimeEnvironment, fn: () => Promise<u
         return;
     }
 
-    const port = await getFreePort();
-    const networkUrl = `http://127.0.0.1:${port}`;
-    hre.config.networks.integratedDevnet.url = networkUrl;
+    const { hostname, port } = new URL(hre.config.starknet.networkUrl);
+    const isPortFree = await isFreePort(parseInt(port));
+    if (!isPortFree) {
```
FabijanC, line 559: Why this check? Couldn't it surprise the user if we just silently change the specified port. If the port is occupied, integrated-devnet spawner will inform the user about that, right?

My understanding was that we should find a free port only if the user didn't provide a URL? Now I'm actually thinking, is it even possible for the user to omit the URL? Can the type extensions file be modified to allow that?

FabijanC (commented):
Should we update any of the tests for this?

FabijanC:
As already discussed, this doesn't have to be a part of release-0.7.0 if we keep URL as a config option.

Currently waiting for response from hardhat: https://github.com/NomicFoundation/hardhat/issues/3221

Nathan-SL:
New issue to followup on hardhat that could resolve this. https://github.com/NomicFoundation/hardhat/issues/3323

PR #206, opened 2022-09-01
Calculate MaxFee Implicit

### LINKED ISSUE #164: Implicitly calculate maxFee
Currently users have to manually run `account.estimateFee` before doing `account.invoke` and explicitly set maxFee as an option of the invoke. Ideally this would be done implicitly, by internally calling fee estimation, and multiplying it with a factor (see which factor others use).

This would be a breaking change since currently the maxFee is implicitly 0 by default.

Nathan-SL:
## Usage related changes

-   Implicitly calculates `account.estimateFee()` inside `account.invoke()` function.
-   Closes #164 

-   NA

## Checklist:

-   [x] Formatted the code
-   [x] Created a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/Shard-Labs/starknet-hardhat-example):https://github.com/Shard-Labs/starknet-hardhat-example/pull/79
 
### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -66,6 +66,11 @@ export abstract class Account {
         calldata?: StringMap,
         options?: InvokeOptions
     ): Promise<InvokeResponse> {
+        const maxFee = await this.estimateFee(toContract, functionName, calldata, options);
+        options = {
+            ...options,
+            maxFee: options?.maxFee || maxFee.amount * BigInt(2)
+        };
         return (
             await this.interact(InteractChoice.INVOKE, toContract, functionName, calldata, options)
         ).toString();


src/account.ts
```diff
@@ -66,6 +66,21 @@ export abstract class Account {
...
+            throw new StarknetPluginError(msg);
+        }
+
+        if (!options?.maxFee) {
+            let overhead = options?.overhead || 0.5;
+            overhead = Math.round((1 + overhead) * 100);
+            maxFee = await this.estimateFee(toContract, functionName, calldata, options);
+            options = {
+                ...options,
+                maxFee: options?.maxFee || (maxFee.amount * BigInt(overhead)) / BigInt(100)
```
FabijanC, line 81: Why again check for `options?.maxFee` ? Isn't the check in L75 enough?

src/account.ts
```diff
@@ -66,6 +66,21 @@ export abstract class Account {
         calldata?: StringMap,
         options?: InvokeOptions
     ): Promise<InvokeResponse> {
+        let maxFee: starknet.FeeEstimation;
+        if (options?.maxFee && options?.overhead) {
+            const msg = "Both maxFee and overhead cannot be specified";
+            throw new StarknetPluginError(msg);
+        }
+
+        if (!options?.maxFee) {
```
FabijanC, line 75: What if maxFee is 0? We should use it in that case and only estimate if it's null or undefined

src/account.ts
```diff
@@ -66,6 +66,21 @@ export abstract class Account {
...
+            let overhead = options?.overhead || 0.5;
```
FabijanC, line 76: What if overhead is 0? We should default to 0.5 only if overhead is null or undefined

scripts/test.sh
```diff
@@ -9,7 +9,7 @@ CONFIG_FILE_NAME="hardhat.config.ts"
 
 # setup example repo
 rm -rf starknet-hardhat-example
-git clone -b release-0.7.0 --single-branch git@github.com:Shard-Labs/starknet-hardhat-example.git
+git clone -b maxfee-inside-invoke --single-branch git@github.com:Shard-Labs/starknet-hardhat-example.git
```
FabijanC, line 12: example repo PR has been merged

FabijanC (changes requested):

src/account.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/206/commits/253aeb445b361f5e6f81c600166c7f916ef2a545

src/account.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/206/commits/253aeb445b361f5e6f81c600166c7f916ef2a545

scripts/test.sh
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/206/commits/253aeb445b361f5e6f81c600166c7f916ef2a545

src/account.ts
```diff
@@ -66,6 +66,21 @@ export abstract class Account {
...
+        let maxFee: starknet.FeeEstimation;
+        if (options?.maxFee && options?.overhead) {
+            const msg = "Both maxFee and overhead cannot be specified";
+            throw new StarknetPluginError(msg);
+        }
+
+        if (options?.maxFee === undefined || options?.maxFee === null) {
+            let overhead = options?.overhead === undefined || options?.overhead === null ? 0.5 : options?.overhead;
+            overhead = Math.round((1 + overhead) * 100);
+            maxFee = await this.estimateFee(toContract, functionName, calldata, options);
```
FabijanC, line 78: I was confused for some time because I didn't realize this maxFee was already defined in L69. I think it would be better to call this `const estimatedFee`.

FabijanC (changes requested):
Can you also update the docs?

src/account.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/206/commits/656599a8198b1684cbc7edc1d9b02d2b9d86a2e6

PR #210, opened 2022-09-14
Fix Docker Connection Timed out

### LINKED ISSUE #209: Error in plugin Starknet: starknet-docker-proxy-20959113038453658 connection timed out!
From what users have said, my guess is that it's a problem with reaching the proxy server inside a docker container on mac host machines.

Nathan-SL:
## Usage related changes

-   Removes network mode from docker args and uses port mapping in place for arm64 machine.
-   Closes #209 

-   NA

## Checklist:

-   [x] Formatted the code
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/starknet-docker-proxy.ts
+++ b/src/starknet-docker-proxy.ts
@@ -33,10 +33,10 @@ export class StarknetDockerProxy extends DockerServer {
         }
 
         const dockerArgs = [...volumes];
-        // Check host machine
-        const isArm64 = process.arch === "arm64";
-        if (isArm64) {
-            this.port = await getFreePort();
+        // Check host os
+        const isDarwin = process.platform === "darwin";
+        if (isDarwin) {
+            this.port = await this.getPort();
             dockerArgs.push("-p", `${this.port}:${this.port}`);
         } else {
             dockerArgs.push("--network", "host");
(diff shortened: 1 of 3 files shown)


src/starknet-docker-proxy.ts
```diff
@@ -32,7 +32,17 @@ export class StarknetDockerProxy extends DockerServer {
             volumes.push("-v", `${mirroredPath}:${mirroredPath}`);
         }
 
-        return [...volumes, "--network", "host"];
+        const dockerArgs = [...volumes];
+        // Check host machine
+        const isArm64 = process.arch === "arm64";
```
FabijanC, line 37: This is an os problem, not an architecture one, the check for mac should be done as in `adaptUrl`

src/starknet-docker-proxy.ts
```diff
@@ -32,7 +32,17 @@ export class StarknetDockerProxy extends DockerServer {
...
+        if (isArm64) {
+            this.port = await getFreePort();
```
FabijanC, line 39: As discussed, there should be only one place where `getFreePort` is called. We could introduce a `getPort` method which returns `this.port` if defined, otherwise calls `getFreePort` before that. The new method would have to return a Promise (i.e. be async).

FabijanC (changes requested):

src/starknet-docker-proxy.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/210/commits/73d4039c6eca8654e2f5d2d306b506f9d213cf1b

src/starknet-docker-proxy.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/210/commits/73d4039c6eca8654e2f5d2d306b506f9d213cf1b

PR #218, opened 2022-09-22
[chore] Update API link in README.md

ivpavici:
## Usage related changes

Docs change, no impact on users

This way, the documentation will stay true even if the line referenced changes in the future commits

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,6 +19,7 @@
 -   [ ] Performed a self-review of the code
 -   [ ] Rebased to the last commit of the target branch (or merged it into my branch)
 -   [ ] Documented the changes
+    -   [ ] If `src/type-extensions.ts` was changed, updated the line number referencing this file in the `## API` section of README.md
 -   [ ] Updated the `test` directory (with a test case consisting of `network.json`, `hardhat.config.ts`, `check.sh`)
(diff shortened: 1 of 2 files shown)


FabijanC:
> This way, the documentation will stay true even if the line referenced changes in the future commits

It will stay the same, but it will stop being true as soon as we add new content to the file, right? We will have to update that link with each new commit that adds new functionality to the `starknet` object. In fact we don't know the hash until we create the commit, so that's also a problem. It's true that currently we also have to manually update when new functionality is added, but we only have to change the line linked, as we are always linking to the correct state of the file.

ivpavici:
> > This way, the documentation will stay true even if the line referenced changes in the future commits
> 
> It will stay the same, but it will stop being true as soon as we add new content to the file, right? We will have to update that link with each new commit that adds new functionality to the `starknet` object. In fact we don't know the hash until we create the commit, so that's also a problem. It's true that currently we also have to manually update when new functionality is added, but we only have to change the line linked, as we are always linking to the correct state of the file.

True, it's kind of a pick your poison situation :D So shall I just edit the line number ?

FabijanC (changes requested):
Alright, just update the line number, but maybe you can also add an entry to the PR template checklist that reminds the developer to update this line.

FabijanC (approved):

PR #274, opened 2022-12-13
 Add timeout to devnet & integrated-devnet requests

### LINKED ISSUE #167: Add timeout to starknet.devnet requests
`starknet.devnet.load` and `starknet.devnet.dump` are reported to hang forever due to the lack of timeout.

Since this would require adding the same timeout config to all the requests, devnet-utils.ts could use a refactor to extract the commonalities.

EDIT: devnet-utils.ts is not the only place where the changes should be applied - also in integrated-devnet classes.

Nathan-SL:
## Usage related changes

-   Adds timeout to requests in `devnet-utils.ts` and `externatl-server.ts`
-   Have 30 seconds as a request timeout (so far, the default axios was 0)
-   Closes #167 

-   NA

## Checklist:

-   [x] Formatted the code

### DIFF
--- a/src/devnet-utils.ts
+++ b/src/devnet-utils.ts
@@ -1,4 +1,5 @@
 import axios, { AxiosResponse, Method } from "axios";
+
 import { StarknetPluginError } from "./starknet-plugin-error";
 import { Devnet, HardhatRuntimeEnvironment } from "hardhat/types";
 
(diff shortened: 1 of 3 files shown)


src/devnet-utils.ts
```diff
@@ -56,94 +57,125 @@ export class DevnetUtils implements Devnet {
...
+        url: string,
+        method: Method,
+        errorMessage?: string,
+        data?: unknown,
+        headers?: any
+    ): Promise<AxiosResponse> {
         try {
-            return await asyncFn();
+            // Make the request
+            const axiosInstance = axios.create({
```
FabijanC, line 69: Could this be a private property?

src/devnet-utils.ts
```diff
@@ -56,94 +57,125 @@ export class DevnetUtils implements Devnet {
         return `${this.hre.starknet.networkConfig.url}`;
     }
 
-    private async withErrorHandler<T>(asyncFn: () => Promise<T>, errorMessage: string) {
+    private async requestHandler(
+        url: string,
+        method: Method,
+        errorMessage?: string,
+        data?: unknown,
+        headers?: any
```
FabijanC, line 65: Where is `headers` used? If used anywhere, could it be a type other than `any`?

src/devnet-utils.ts
```diff
@@ -56,94 +57,125 @@ export class DevnetUtils implements Devnet {
...
-        return this.withErrorHandler<FlushResponse>(async () => {
-            const response = await axios.post<FlushResponse>(`${this.endpoint}/postman/flush`);
-            return response.data;
-        }, "Request failed. Make sure your network has the /postman endpoint");
+        const response = await this.requestHandler(
+            "/postman/flush",
+            "POST",
+            "Request failed. Make sure your network has the /flush endpoint"
+        );
+        return response.data as FlushResponse;
```
FabijanC, line 97: Instead of casting here, could `requestHandler` be parametrized with a generic `T` (kind of like it used to be)

src/devnet-utils.ts
```diff
@@ -56,94 +57,125 @@ export class DevnetUtils implements Devnet {
         return `${this.hre.starknet.networkConfig.url}`;
     }
 
-    private async withErrorHandler<T>(asyncFn: () => Promise<T>, errorMessage: string) {
+    private async requestHandler(
+        url: string,
+        method: Method,
+        errorMessage?: string,
```
FabijanC, line 63: Instead of passing in the error message, I would suggest constructing it like this just before throwing it:
```typescript
`Request failed: Could not ${method} ${url}. This is a Devnet-specific functionality. Make sure you really want to interact with Devnet and that it is running and available at ${this.endpoint}`
```

FabijanC (changes requested):

src/devnet-utils.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/274/commits/cad87ed19051704fb1e0dc240df16f8583ab878e

src/devnet-utils.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/274/commits/cad87ed19051704fb1e0dc240df16f8583ab878e

src/devnet-utils.ts
Nathan-SL: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/274/commits/cad87ed19051704fb1e0dc240df16f8583ab878e

src/devnet-utils.ts
FabijanC: Marking as resolved as I see this was also covered in https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/274/commits/cad87ed19051704fb1e0dc240df16f8583ab878e.

FabijanC (approved):

PR #295, opened 2023-01-18
231 amarna

### LINKED ISSUE #231: Add Amarna static-analyzer and linter for Cairo
Add a command (or integrate somehow else) for users to call the amarna static-analyzer and linter for their cairo contracts to detect potential problems

https://github.com/crytic/amarna

shramee:
## Usage related changes

Adds `npx hardhat amarna` to run Amarna in a docker container and analyse and lint Cairo files in the project and output `out.sarif` file.

- New  `AmarnaDocker` extending `DockerServer`
- New task amarna to runs Amarna via AmarnaDocker.

Fixes #231 Add Amarna static analyser

-   Change y.
-   ...

## Checklist:

-   [x] Formatted the code
    -   Shard-Labs/starknet-hardhat-example#91


src/external-server/docker-amarna.ts
```diff
@@ -0,0 +1,42 @@
...
+    protected async getDockerArgs(): Promise<string[]> {
+        // To access the files on host machine from inside the container, proper mounting has to be done.
+        const volumes = ["-v", `${this.path}:/src`];
+        const dockerArgs = [...volumes];
+        if (this.useShell) {
+            // Run ./amarna.sh file for custom args
+            if (fs.existsSync(`${this.path}/amarna.sh`)) {
+                dockerArgs.push("--entrypoint", "./amarna.sh");
+            } else {
+                console.log(
```
FabijanC, line 24: Should it be `console.warn` instead?

src/external-server/docker-amarna.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as fs from "fs";
+
+export class AmarnaDocker extends DockerServer {
+    useShell = false;
+    /**
+     * @param image the Docker image to be used for running the container
+     * @param cairoPaths the paths specified in hardhat config cairoPaths
+     */
+    constructor(image: Image, private path: string) {
+        super(image, "127.0.0.1", null, "", "starknet-docker-proxy");
```
FabijanC, line 12: It should probably use a different container name instead of `starknet-docker-proxy` since `StarknetDockerProxy` uses that one. Saying this to prevent name conflicts, although not sure if the two containers can be run at the same time.

src/external-server/docker-amarna.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    protected async getDockerArgs(): Promise<string[]> {
+        // To access the files on host machine from inside the container, proper mounting has to be done.
+        const volumes = ["-v", `${this.path}:/src`];
+        const dockerArgs = [...volumes];
+        if (this.useShell) {
+            // Run ./amarna.sh file for custom args
+            if (fs.existsSync(`${this.path}/amarna.sh`)) {
```
FabijanC, line 21: Is this some amarna native feature? If so, where is it documented? If not, what is it?

FabijanC (changes requested):
Thanks for dealing with this task!

Could this feature be tested somehow?

The docs should be updated (hence the `Documented the changes` point in the TODO list).

src/external-server/docker-amarna.ts
shramee: We use `-rm`, but should totally rename it!

src/external-server/docker-amarna.ts
shramee: Yes, definitely!

src/external-server/docker-amarna.ts
shramee: It's in case some projs wants to use some of the ton of config options Amarna provides.

FabijanC: Ok, sounds useful, let's keep it as it is, but documented!

shramee: Aye cap!

FabijanC (commented):
Looking good after the changes, but what about testing this feature? Something like https://github.com/Shard-Labs/starknet-hardhat-plugin/blob/master/test/general-tests/expect-error-on-compile/check.ts

Or at least some light script which at least runs the new command.

test/general-tests/amarna/check.ts
```diff
@@ -0,0 +1,3 @@
+import { hardhatStarknetTest } from "../../utils/cli-functions";
+
+hardhatStarknetTest("--no-compile test/amarna.test.ts".split(" "));
```
FabijanC, line 3: Instead of executing `npx hardhat test` on a script from the example repo, why not just have all that logic inside this `check.ts` script? I.e. it seems to me that this could be tested without the separate example-repo PR (https://github.com/Shard-Labs/starknet-hardhat-example/pull/91). That's what I was aiming at with this comment: https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/295#pullrequestreview-1263803398

FabijanC (changes requested):

test/general-tests/amarna/check.ts
shramee: Aah true.
🤦‍♂️

shramee: Aah, now I remember... It needs Cairo files in the root of the project tho. So it seemed to make sense to put in the example plug-in.

But we could move it here too.

FabijanC: The tests are executed from the root of the project, so you can count on the source files to be present.

Check this test case for example: https://github.com/Shard-Labs/starknet-hardhat-plugin/tree/master/test/general-tests/expect-error-on-compile

It needs its own cairo file so it first copies it into the example project (in which everything is executed).

shramee: Aah yeah, it does. Fantastic!

shramee: Hi @FabijanC,

This is a great suggestion! Now we can call `npx hardhat amarna` instead of using `AmarnaDocker` internally.
Which also tests if the class is properly bound to the task. Brill!

test/utils/utils.ts
```diff
@@ -19,6 +19,14 @@ export function assertContains(output: string, pattern: string) {
     }
 }
 
+export function assertEquals(actual: string, expected: string) {
```
FabijanC, line 22: Why not use the `assertEqual` function already defined in this file?

FabijanC (changes requested):
Thanks for adding the comments to the test!

test/utils/utils.ts
shramee: Aah, shultz...
That was mighty retarded... 🤦

shramee: Fixed

FabijanC (approved):

PR #296, opened 2023-01-21
290 Tests in docker

### LINKED ISSUE #290: Make running tests easier
Running tests involves a lot of friction this makes the dev cycle slower. Here are some ideas,

## 1. Dockerize the tests
A priviledged Alpine/Debian container to run the tests.
1. One time setup (the image and container)
```sh
# Build the image
docker build -t starknet-hardhat-plugin-tester .

# Uncomment the line below to remove the old container
# docker rm starknet-hardhat-plugin-tester

# Create a new container
docker create --name starknet-hardhat-plugin-tester starknet-hardhat-plugin-tester
```

2. Running tests.
```sh
# Copy the latest plugin
docker cp . starknet-hardhat-plugin-tester:/node

# The new image will have the files
docker run -it starknet-hardhat-plugin-tester /bin/sh
```

3. Then inside container we just do `npm i` and `npm run test-general-tests` etc. to run tests.

## 2. Mock devnet etc dependencies

> We might have discussed this before and decided to go the other way, let me know what you guys think.

An Axios mock would be really neat. This would allows tests to run much faster and be great for TDD.
So for example a test for devnet.mint #285 would just check if the `mint` method makes a request to `/mint` endpoint with the correct data.
The drawback would be if devnet changes the way request to mint tokens is handled the plugin might be broken while the tests still pass. However with devnet instance locked to a specific version we should be all good.

shramee:
## Usage related changes

- Users are unaffected.

#### Executing tests in docker

You can use docker tester image. The image includes testing environment set up with everything required to run the tests. The spawned container has access to docker daemon so it can also run the tests that require spawning Devnet/Cairo CLI etc containers.

```
npm run docker-tester
```

This is take you in the container shell. Once in the container, you can use the `test-` scripts defined in `package.json` in an isolated environment.

Fixes #290

## Checklist:

-   [x] Formatted the code
    -   < EXAMPLE_REPO_PR_URL > 


shramee:
Sorry, this needs more testing.

shramee:
Add a ton of docs, let me know if it's a bit much and some of it is unnecessary...

--------------------------

## Dockerized testing environment

We have added a Docker in Docker, called DinD, setup to run the tests that need to use Docker from within a Docker container.
This allows running tests easily without changing your development environment. Read about running tests in [Executing tests in docker](#executing-tests-in-docker) section. If you are not adding **Docker container volumes** you can skip this section.

Running tests inside DinD requires some additional setup/parsing when adding new Docker container volumes to map container paths.

### Adding new docker container

If you are adding a new docker container with volumes please refer to the next section, [Adding volumes to a docker container](#adding-volumes-to-a-docker-container), to properly parse host paths for Docker in Docker to work.

### Adding volumes to a docker container

For volume mounts, Use `getVolumeHostPathFilter` function from `./external-server/external-server` to get the filter function to prepare paths for DinD if required.

Here's an example,

```js
// import ...
import { getVolumeHostPathFilter } from "./external-server/external-server";

// A bunch of code...

const volumeHostPathFilter = getVolumeHostPathFilter();
// Docker volume mounting args
const volumes = [
    "-v",
    `${volumeHostPathFilter(pathInHost)}:${pathInContainer}`,
    "-v",
    `${volumeHostPathFilter(pathInHost2)}:${pathInContainer2}`
];

// Other bunch of code...
```

We get the filter function from `getVolumeHostPathFilter()`. Then `host` parts for every volume are passed filtered with `volumeHostPathFilter` function.

## Testing

### ...

### Executing tests in docker

You can use docker tester image. The image includes testing environment set up with everything required to run the tests. The spawned container has access to docker daemon so it can also run the tests that require spawning Devnet/Cairo CLI etc containers. Read more in [Dockerized testing environment](#dockerized-testing-environment) section.

```
npm run docker-tester
```

This is take you in the container shell. Once in the container, you can use the `test-` scripts defined in `package.json` in an isolated environment.

### ...

scripts/ensure-python.sh
```diff
@@ -7,6 +7,7 @@ set -eu
 PY_VERSION=3.8.9
 
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-    which "/opt/circleci/.pyenv/versions/$PY_VERSION/bin/python" || pyenv install "$PY_VERSION"
+		# Install version PY_VERSION, but skip if already installed.
+    which "/opt/circleci/.pyenv/versions/$PY_VERSION/bin/python" || pyenv install "$PY_VERSION" -s
```
shramee, line 11: `-s` flag skips if `pyenv` version is already installed.
https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-install
>  -s/--skip-existing    Skip the installation if the version appears to be installed already

scripts/install-devnet.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eu
 
-pip3 install "starknet-devnet==$STARKNET_DEVNET"
+# Get default from config.json file
+STARKNET_DEVNET_DEFAULT=$(node -e "console.log(require('../config.json').STARKNET_DEVNET)")
```
shramee, line 6: Sets default for `STARKNET_DEVNET` env var from `config.json`

scripts/install-devnet.sh
```diff
@@ -2,6 +2,9 @@
...
+
+pip3 install "starknet-devnet==${STARKNET_DEVNET:=$STARKNET_DEVNET_DEFAULT}" | { grep --invert-match 'Requirement already satisfied:' || :; }
```
shramee, line 8: 1. Uses STARKNET_DEVNET_DEFAULT fallback
2. grep removes `Requirement already satisfied` logs.
3. `|| :;` part, which is like `or nothing;`, prevents grep from `exit`ing the process.

scripts/setup-venv.sh
```diff
@@ -13,7 +13,7 @@ echo "python version: $(python --version)"
 if [[ "$OSTYPE" == "darwin"* ]]; then
     export HOMEBREW_NO_INSTALL_CLEANUP=1
     brew ls --versions gmp || brew install gmp
-    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip3 install fastecdsa
+    CFLAGS=-I`brew --prefix gmp`/include LDFLAGS=-L`brew --prefix gmp`/lib pip3 install fastecdsa  | { grep --invert-match 'Requirement already satisfied:' || :; }
```
shramee, line 16: Same as https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/296/files#r1115318562
Strips `Requirement already satisfied` logs and stops grep from exiting the process.

scripts/test.sh
```diff
@@ -10,12 +10,12 @@ CONFIG_FILE_NAME="hardhat.config.ts"
 # setup example repo
 rm -rf starknet-hardhat-example
 EXAMPLE_REPO_BRANCH="plugin"
-if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin" ]]; then
+if [[ "${CIRCLE_BRANCH:=}" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin" ]]; then
```
shramee, line 13: Empty default from `CIRCLE_BRANCH`

FabijanC (changes requested):
Your PRs seems to introduce changes that are already on master. Perhaps merging master into your branch would help (also there are conflicts with the target branch). It should be easier to review after that.

I haven't fully reviewed, but if I understand correctly, some parts of the coding (non-testing code) are now working differently just to enable testing with DinD?

shramee:
Closed in favour of #331

PR #302, opened 2023-01-25
231 amarna ci issue

shramee:
## Usage related changes

None

Fixes Amarna CI test issues.

## Checklist:

-   [x] Formatted the code
    -   < EXAMPLE_REPO_PR_URL > 


shramee:
Continues #295

shramee:
@FabijanC Ready to have a look now.

src/external-server/docker-amarna.ts
```diff
@@ -33,8 +35,30 @@ export class AmarnaDocker extends DockerServer {
 
     public async run(args: { script?: boolean }) {
         this.useShell = !!args.script;
+        const formattedImage = `${this.image.repository}:${this.image.tag}`;
```
FabijanC, line 38: The `pullImage` method was made protected but isn't used, although this looks like a use case for it.

FabijanC (changes requested):
While reviewing this, I realized we've probably been adding Amarna support the wrong way. 

The idea of the `DockerServer` abstract class is to support using Docker containers through a server. That is useful if the container needs to do time-consuming things before doing the actual useful work (e.g. importing the StarkNet compiler before compiling). Also, that is applicable only if Hardhat is expected to use the Docker server multiple times while the Hardhat process is alive. That is why `DockerServer` wasn't really suitable for `docker run -it` operations.

It wasn't always like that; we used to rely on `hardhat-docker` (an official plugin for Docker interaction in Hardhat), and that might just be what we'd need here.

I would say the e2e test is ok (once my suggested change is implemented), so that part can stay.

Let me know if anything needs more clarification or if we should discuss this on a call.

test/general-tests/amarna/check.ts
```diff
@@ -20,21 +20,17 @@ fs.existsSync("./out.sarif") && fs.unlinkSync("./out.sarif");
 
 console.log("Running amarna");
 
-spawnSync("npx hardhat amarna");
+const output = spawnSync("npx", ["hardhat", "amarna"], { encoding: "utf-8" });
+output.stdout && console.log(output.stdout);
```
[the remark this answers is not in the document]
shramee: So we have a better visibility of what's going on and if it fails.

src/external-server/docker-amarna.ts
shramee: Aye, but it doesn't do it in sync, so we can't use it :(

FabijanC: I don't understand, why `await this.pullImage(...)` be used?

FabijanC:
If it's ready for another round of review, I can do it, but tomorrow.

src/external-server/docker-amarna.ts
```diff
@@ -1,27 +1,24 @@
...
+    constructor(private image: Image, private path: string) {
+        this.container = "amarna-container-" + Math.random();
     }
 
-    protected async getDockerArgs(): Promise<string[]> {
-        // To access the files on host machine from inside the container, proper mounting has to be done.
-        const volumes = ["-v", `${this.path}:/src`];
-        const dockerArgs = [...volumes];
+    protected getDockerArgs(): string[] {
+        let cmd = ["amarna", ".", "-o", "out.sarif"];
```
FabijanC, line 17: I'd also use slicing to remove `"0."` from the name, as we do in `DockerServer`:
```js
Math.random().toString().slice(2);
```

src/external-server/docker-amarna.ts
```diff
@@ -30,38 +27,29 @@ export class AmarnaDocker extends DockerServer {
...
-    protected async getContainerArgs(): Promise<string[]> {
-        return [];
+        console.log("Running amarna, this may take a while.");
+        const { stdout, stderr } = await this.docker.runContainer(this.image, cmd, {
+            binds: {
+                [path]: "/src"
+            }
+        });
+        // Output the output/error for user to review.
+        console.log(stdout.toString(), stderr.toString());
```
FabijanC, line 53: Why not log stdout to stdout and stderr to stderr?

FabijanC (changes requested):
It should be documented that this feature is only available if the user has docker.

src/external-server/docker-amarna.ts
```diff
@@ -28,16 +27,29 @@ export class AmarnaDocker extends DockerServer {
...
+            await this.docker.pullImage(this.image);
+        }
+        const cmd = this.getDockerArgs();
 
-    protected async getContainerArgs(): Promise<string[]> {
-        return [];
+        console.log("Running amarna, this may take a while.");
+        const { stdout, stderr } = await this.docker.runContainer(this.image, cmd, {
+            binds: {
+                [path]: "/src"
```
FabijanC, line 49: Why doesn't it need to bind other paths as well? Like source paths?

shramee:
@FabijanC It's ready to have a look 😉 but not sure I understand **source paths**.

FabijanC (changes requested):
@shramee And let's not forget to update the docs to say that this feature is only available if the user has Docker installed.

src/external-server/docker-amarna.ts
FabijanC: Resolving this conversation, continued in https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/302#discussion_r1090316707

shramee:
Added the docs and Cairo paths future proofing 😉

FabijanC (changes requested):
Should have tried this long before, but this is what I got after running `npx hardhat amarna`:
```
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
exec /usr/local/bin/amarna: exec format error
```

FabijanC:
Also, linter is reporting an unused-import error

FabijanC:
@shramee The warning I mentioned in [this comment](https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/302#pullrequestreview-1278754963) - that's the only thing that gets printed and no output file is generated. That's the same thing that happened on CircleCI, the image platform was incompatible. I would expect it to fail in that case, but looks like it didn't.

Also, did we see this warning in the CircleCI logs? If not, why?

shramee:
The error for platform shouldn't affect the functioning, however I don't think I've seen the exec error. Let me know if you see that again.

shramee:
I amended the last commit to remove `out.sarif` file. You might need to pull again...

FabijanC:
> The error for platform shouldn't affect the functioning, however I don't think I've seen the exec error. Let me know if you see that again.

It shouldn't but it does. For me it just prints the warning (it's not even an error, it exits with code 0) and doesn't produce any file outputl

shramee:
I think I need to deploy a multi arch image... To support both arm/v8 and amd64

shramee:
Also, I think that `exec` error might be the one halting the everything, do we have any more info about it?

shramee:
Yeep, I think we need multi-arch 💪

FabijanC:
> Yeep, I think we need multi-arch muscle

Will you be able to take care of the building?

shramee:
Aye ser, leave it to me.
But might need a while as I'm also doing a HackerBuilding project. That okay?

FabijanC:
> Aye ser, leave it to me. But might need a while as I'm also doing a HackerBuilding project. That okay?

Ok! But let's try to resolve it at least next week then. We might then temporarily disable the Amarna test until this PR is merged (in which case this PR could then also enable back the disabled test).

shramee:
Multi arch image is deployed :fire:

FabijanC (changes requested):
Alright, it's working now with the multiarch image! I ran it on my computer and it generated `out.sarif`. I think it would be useful to print to the user where the output file has been generated.

FabijanC:
@shramee As announced, due to temporarily disabling the Amarna test on `master`, this PR now has a conflict with the target branch. You can resolve it together with addressing the remaining change requests.

shramee:
Woop Woop! Awesome news!
Aye indicating the generated file path sounds like an awesome idea. 🥳

FabijanC:
@shramee Could we have this merged within the next hour or so? Planning to release a new version.

shramee:
Sorry, I was at Jerusalem today, just seen now.

src/external-server/docker-amarna.ts
```diff
@@ -28,16 +36,75 @@ export class AmarnaDocker extends DockerServer {
...
+
         console.log("Running amarna, this may take a while.");
-        await this.spawnChildProcess();
-    }
 
-    protected async getContainerArgs(): Promise<string[]> {
-        return [];
+        const result = spawnSync("docker", ["run", ...dockerArgs]);
+
+        if (!this.useShell) {
```
FabijanC, line 102: Why does this depend on `useShell`? What exactly is the purpose of `useShell`?

FabijanC (commented):
Btw, it turns out external contributors can trigger our CircleCI workflow. I enabled that feature yesterday, so your last two commits have green checks indicating all tests have passed, including the Amarna test: [test log link](https://app.circleci.com/pipelines/github/Shard-Labs/starknet-hardhat-plugin/1587/workflows/75ea4e99-6248-4af6-b4d5-917e7852132f/jobs/7029?invite=true#step-111-473).

I just kindly ask you not to use it recklessly in future PRs. You can put `[skip ci]` in your commit message to prevent workflow execution, which is recommendable if you are only improving docs or doing something very minor.

PR #306, opened 2023-02-02
Get account balance

### LINKED ISSUE #300: Create utility function for fetching account balance
The function would rely on polling the contract at address `0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7`. This is assuming token contracts on all networks have the `balanceOf` function. The plugin would need to have the ABI of the token contract (other compilation artifact as well?).

TsBauer:
## Usage related changes

-   Adding a utility function for fetching balance of a smart contract

-   N/A

## Checklist:

-   [x] Formatted the code
-   [x] No linter errors

-----------------------------------------------------------

OLD:

This PR aims to resolve #300 
Everything is straight forward. We added the ERC20 abi and a `getAccountBalance()` function to the Account class.

### Open question

Right now, the return value is a big number like this: `{ balance: { low: 441031929613452n, high: 0n } }`
Should we convert this to a plain `number`?

Cheers


src/account.ts
```diff
@@ -106,6 +106,21 @@ export abstract class Account {
         return this.starknetContract.address;
     }
 
+    async getAccountBalance(): Promise<StringMap> {
```
FabijanC, line 109: Instead of having a method, I think it would be better to have a utility function `getBalance` which accepts an address and returns the corresponding balance. When I say utility function, I mean something like [starknet.getTransaction](https://shard-labs.github.io/starknet-hardhat-plugin/docs/intro/#transaction-information-and-receipt-with-events) that doesn't belong to an instantiated class.

src/account.ts
```diff
@@ -106,6 +106,21 @@ export abstract class Account {
...
+
+        const hre = await import("hardhat");
+        const ethContract = new StarknetContract({
+            abiPath: abiPath,
+            hre: hre
+        });
+        ethContract.address = "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7";
+
+        return (
+            await ethContract.call("balanceOf", {account: this.starknetContract.address})
```
FabijanC, line 120: The value should be extracted and converted to Bigint (this answers the "open question" from the PR description). Even though starknet-hardhat-plugin relies on an older version of starknet.js, even that one should have a utility function for this conversion. We prefer Bigint over Number as it preserves precision.

src/account.ts
```diff
@@ -106,6 +106,21 @@ export abstract class Account {
         return this.starknetContract.address;
     }
 
+    async getAccountBalance(): Promise<StringMap> {
+        const abiPath = process.cwd() + "/contract-artifacts/ERC20/ERC20_abi.json";
```
FabijanC, line 110: Why not use `hre.starknet.getContractFactory`?

src/account.ts
```diff
@@ -106,6 +106,21 @@ export abstract class Account {
...
+
+        const hre = await import("hardhat");
+        const ethContract = new StarknetContract({
+            abiPath: abiPath,
+            hre: hre
+        });
+        ethContract.address = "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7";
```
FabijanC, line 117: This could be a constant in `src/constants.ts`

FabijanC (changes requested):
I see you removed the content of the PR template. It contains useful points such as:
- documentation reminder
- testing instructions (also documented in the [dev docs](https://shard-labs.github.io/starknet-hardhat-plugin/docs/dev/#testing))

src/account.ts
TsBauer: Agree. This has been refactored accordingly.

src/account.ts
TsBauer: Alright, we return now BigInt. Is there a more direct way of conversion than converting the result first to BigNumber, then to string and finally BigInt?

src/account.ts
TsBauer: There's no particular good reason. This has been refactored👍

src/account.ts
TsBauer: Refactored accordingly 👍

TsBauer:
We have now fully implemented your feedback. Two questions remain:
1. Do you expect further documentation?
2. Our tests are passing. But others are failing even without any from this branch. I'm hesitant on making a checkmark on `All tests are passing`. But I might be doing something wrong here.

src/account.ts
FabijanC: Nice approach, can't think of anything better.

FabijanC (commented):
I'm usually skeptical of new user contributions due to the complicated nature of testing, but you nailed this one :partying_face:.

To answer your questions
> 1. Do you expect further documentation?

The JSdoc you wrote looks good. Honestly I thought our `www/docs/intro.md` document already contained a section dedicated to something similar (best we have is a paragraph about account funding), so you don't have to add anything there, we'll probably deal with that in the near future.

> 2. Our tests are passing. But others are failing even without any from this branch. I'm hesitant on making a checkmark on `All tests are passing`. But I might be doing something wrong here.

Is it just one test? We've recently been having a problem with the Amarna test, so it's ok if it's just that test failing. That test should pass after https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/302 is merged.

src/extend-utils.ts
```diff
@@ -182,3 +185,17 @@ export async function getNonceUtil(
...
     return parseInt(executed.stdout.toString());
 }
+
+export async function getBalanceUtil(
+    address: string,
+    hre: HardhatRuntimeEnvironment
+): Promise<BigInt> {
+    const contractPath = handleInternalContractArtifacts("Token", "ERC20", "", hre);
+    const contractFactory = await hre.starknet.getContractFactory(contractPath);
+    const ethContract = contractFactory.getContractAt(ETH_ADDRESS);
```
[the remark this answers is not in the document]
TsBauer: Measured with `performance.now()` line 193 yields 0.10141658782958984 ms on a Macbook Air M1. It certainly feels instant to me but it's no problem to relocate this line. If so, would you prefer having it in `constants.ts`?

TsBauer:
> Is it just one test? We've recently been having a problem with the Amarna test, so it's ok if it's just that test failing. That test should pass after https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/302 is merged.

We've been running the suite again and 88 tests are passing and 17 are failing. Our tests are passing and to the best of our knowledge there's no influence of our changes to the failed tests.

FabijanC:
> > Is it just one test? We've recently been having a problem with the Amarna test, so it's ok if it's just that test failing. That test should pass after #302 is merged.
> 
> We've been running the suite again and 88 tests are passing and 17 are failing. Our tests are passing and to the best of our knowledge there's no influence of our changes to the failed tests.

Then there's probably something with your environment setup, but I'd say it's probably safe to merge, I'll deal with this later in the day.

FabijanC (approved):
Approving this, will merge to our side branch to see how the tests go. On that branch I'll modify the example-repo branch to `plugin`.

FabijanC:
Superseded by #310

PR #309, opened 2023-02-07
Fix: typos

omahs:
Fix: typos


FabijanC (approved):
Thanks for being a careful reader.

PR #322, opened 2023-02-22
Allow axios timeout configuration via hardhat config

### LINKED ISSUE #318: Allow request timeout customization
As described here: https://discord.com/channels/793094838509764618/1073370808086437899

We would need to see where we use the currently hardcoded constant `REQUEST_TIMEOUT` defined in `src/constants.ts`.

notV4l:
## Usage related changes

- Close #318 
- allow axios timeout configuration via hardhat.config.js

-  new parameter in hardhat.config.js

```
starknet: {
    requestTimeout: 90_000,
}
```

## Checklist:

-   [x] Formatted the code
-   [ ] Rebased to the last commit of the target branch (or merged it into my branch)
    -   < EXAMPLE_REPO_PR_URL > 


src/external-server/external-server.ts
```diff
@@ -163,7 +164,7 @@ export abstract class ExternalServer {
 
         try {
             const response = await axios.post<T>(this.url, data, {
-                timeout: REQUEST_TIMEOUT,
+                timeout: this.hre.config.axios.timeout,
```
FabijanC, line 167: Can't you dynamically import the hardhat runtime environment? Rather than having to modify all the constructors and propagate it.

Check other instances of doing `await import("hardhat")`

src/index.ts
```diff
@@ -83,6 +83,18 @@ extendConfig((config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) =>
     }
 });
 
+// copy all user-defined axios settings
+extendConfig((config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) => {
+    if (userConfig.axios) {
+        config.axios = JSON.parse(JSON.stringify(userConfig.axios));
+    }
+    if (!config.axios) {
+        config.axios = {
```
FabijanC, line 92: Two things:
- Giving this an implementation specific name is not a good idea, it should be more abstract/generic
- To preserve namespace independence between various hardhat plugins, I'd recommend making this piece of settings a part of the `starknet` config object.

That being said, perhaps we could use an option called `httpTimeout` or `requestTimeout`.

FabijanC (changes requested):
You should expand `test/configuration-tests`. No additions would be required to the starknet-hardhat-example repo.

You could introduce two new test cases:
    - one with an extremely small timeout (1 ms) and expected to fail
    - another with a sufficient timeout (e.g. 30 seconds) and expected to pass

src/external-server/external-server.ts
notV4l: indeed XD

src/index.ts
notV4l: changed to starknet.requestTimeout

FabijanC (approved):

PR #327, opened 2023-02-27
Docker Desktop support

### LINKED ISSUE #321: WSL issues
WSL might not be completely compatible with the plugin.

Can host.docker.internal be used on Linux? Handling localhost uniformly would simplify the code and potentially prevent issues.
Possible solution, didn't get it to work locally on my Linux machine: https://stackoverflow.com/questions/30109037/how-can-i-forward-localhost-port-on-my-container-to-localhost-on-my-host

notV4l:
## Usage related changes

- Close #321 

## Checklist:

-   [x] Formatted the code
-   [ ] Rebased to the last commit of the target branch (or merged it into my branch)
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/external-server/docker-server.ts
+++ b/src/external-server/docker-server.ts
@@ -37,6 +37,7 @@ export abstract class DockerServer extends ExternalServer {
         await this.pullImage();
 
         const formattedImage = `${this.image.repository}:${this.image.tag}`;
+
         const args = [
             "run",
             "--rm",
@@ -46,7 +47,13 @@ export abstract class DockerServer extends ExternalServer {
             formattedImage,
             ...(await this.getContainerArgs())
         ];
+
         return spawn("docker", args);
+    }
+
+    public isDockerDesktop(): boolean {
+        const res = spawnSync("docker", ["system", "info"], { encoding: "utf8" });
+        return res.stdout.includes("Operating System: Docker Desktop");
     }
 
     /**
--- a/src/external-server/docker-server.ts
+++ b/src/external-server/docker-server.ts
@@ -51,11 +51,6 @@ export abstract class DockerServer extends ExternalServer {
         return spawn("docker", args);
     }
 
-    public isDockerDesktop(): boolean {
-        const res = spawnSync("docker", ["system", "info"], { encoding: "utf8" });
-        return res.stdout.includes("Operating System: Docker Desktop");
-    }
-
     /**
      * CLI arguments passed to the `docker` command.
      */
--- a/src/external-server/external-server.ts
+++ b/src/external-server/external-server.ts
@@ -72,7 +72,8 @@ export abstract class ExternalServer {
      */
     private getIsDockerDesktop(): boolean {
         const res = spawnSync("docker", ["system", "info"], { encoding: "utf8" });
-        return res?.stdout?.includes("Operating System: Docker Desktop");
+        console.log(res);
+        return res.stdout?.includes("Operating System: Docker Desktop");
     }
 
     public get url() {
--- a/src/external-server/external-server.ts
+++ b/src/external-server/external-server.ts
@@ -1,11 +1,10 @@
 import axios from "axios";
 import net from "net";
-import { ChildProcess } from "child_process";
+import { ChildProcess, spawnSync } from "child_process";
 import { StarknetPluginError } from "../starknet-plugin-error";
 import { IntegratedDevnetLogger } from "./integrated-devnet-logger";
 import { StringMap } from "../types";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
-import { isDockerDesktop } from "../utils";
 
 function sleep(amountMillis: number): Promise<void> {
     return new Promise((resolve) => {
@@ -63,9 +62,17 @@ export abstract class ExternalServer {
 
     public get isDockerDesktop(): boolean {
         if (this._isDockerDesktop === null) {
-            this._isDockerDesktop = isDockerDesktop();
+            this._isDockerDesktop = this.getIsDockerDesktop();
         }
         return this._isDockerDesktop;
+    }
+
+    /**
+     * Check if docker is Docker Desktop
+     */
+    private getIsDockerDesktop(): boolean {
+        const res = spawnSync("docker", ["system", "info"], { encoding: "utf8" });
+        return res?.stdout?.includes("Operating System: Docker Desktop");
     }
 
     public get url() {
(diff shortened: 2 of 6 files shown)


src/external-server/docker-server.ts
```diff
@@ -46,9 +47,15 @@ export abstract class DockerServer extends ExternalServer {
             formattedImage,
             ...(await this.getContainerArgs())
         ];
+
         return spawn("docker", args);
     }
 
+    public isDockerDesktop(): boolean {
```
FabijanC, line 54: I would implement this as a cached [JS getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) (does it have to be public?) which first checks a private property to see if it was already executed.

src/external-server/external-server.ts
```diff
@@ -149,6 +149,7 @@ export abstract class ExternalServer {
     }
 
     private async isServerAlive() {
+        if (this.port === null) return false;
```
FabijanC, line 152: Why is this necessary?

src/utils.ts
```diff
@@ -53,14 +53,13 @@ export function adaptLog(msg: string): string {
...
-const MACOS_PLATFORM = "darwin";
 /**
- * Adapts `url` by replacing localhost and 127.0.0.1 with `host.internal.docker`
+ * Adapts `url` by replacing localhost and 127.0.0.1 with `host.docker.internal`
  * @param url string representing the url to be adapted
  * @returns adapted url
  */
-export function adaptUrl(url: string): string {
-    if (process.platform === MACOS_PLATFORM) {
+export function adaptUrl(url: string, isDockerDesktop: boolean): string {
```
FabijanC, line 61: Instead of having to provide the boolean argument, could this `adaptUrl` function find out by itself if the Docker is indeed a desktop one?

FabijanC (changes requested):

src/external-server/docker-server.ts
notV4l: its only executed once when spawning docker container

src/external-server/external-server.ts
notV4l: starknet-docker-proxy is initialized with null port, it avoid one request with null port

Nathan-SL (approved):

src/external-server/external-server.ts
```diff
@@ -57,6 +59,11 @@ export abstract class ExternalServer {
         protected stderr?: string
     ) {
         ExternalServer.cleanupFns.push(this.cleanup.bind(this));
+        this._isDockerDesktop = isDockerDesktop();
+    }
+
+    public get isDockerDesktop(): boolean {
```
FabijanC, line 65: I didn't mean the getter to just return the value of a private property, but also set the value of it so that it doesn't have to be set elsewhere. And so that this getter is the only place responsible for checking.

src/utils.ts
```diff
@@ -77,6 +77,13 @@ export function adaptUrl(url: string): string {
 
     return url;
 }
+/**
+ * Check if docker is Docker Desktop
+ */
+export function isDockerDesktop(): boolean {
+    const res = spawnSync("docker", ["system", "info"], { encoding: "utf8" });
+    return res?.stdout?.includes("Operating System: Docker Desktop");
```
FabijanC, line 85: Why the question marks? Shouldn't these properties always exist? When I earlier said that this looks flaky, I meant that we are depending on the output format of the `docker system info command`, which might change or not be the same on all platforms.

FabijanC (changes requested):

src/utils.ts
notV4l: i guess this logic was in utils.ts to make it accessible from anywhere, that's also why i moved isDockerDesktop function to utils.ts

src/starknet-docker-proxy.ts
```diff
@@ -28,8 +28,10 @@ export class StarknetDockerProxy extends DockerServer {
     }
 
     protected async getDockerArgs(): Promise<string[]> {
-        // To access the files on host machine from inside the container, proper mounting has to be done.
+        // get a valid free port
+        this.port = await this.getPort();
```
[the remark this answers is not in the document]
notV4l: cause StarknetDockerProxy is initialized with port = null

src/starknet-wrappers.ts
```diff
@@ -446,14 +449,16 @@ export class DockerWrapper extends StarknetWrapper {
         cairoPaths: string[],
         hre: HardhatRuntimeEnvironment
     ) {
-        super(new StarknetDockerProxy(image, rootPath, accountPaths, cairoPaths), hre);
+        const externalServer = new StarknetDockerProxy(image, rootPath, accountPaths, cairoPaths);
+        super(externalServer, hre);
+        this.externalServer = externalServer;
```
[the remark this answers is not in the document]
notV4l: indeed

src/starknet-docker-proxy.ts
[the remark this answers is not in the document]
FabijanC: And it's initialized with null to save one request?

src/utils.ts
FabijanC: `adaptUrl` was in utils because it really was used in multiple places. Now it's used in only one place (after a refactor was done a few months ago).

src/utils.ts
notV4l: this morning i had to add question marks to make tests pass while it was ok yesterday :|

src/utils.ts
notV4l: refactored

src/starknet-wrappers.ts
```diff
@@ -486,10 +502,6 @@ export class VenvWrapper extends StarknetWrapper {
         super(new StarknetVenvProxy(pythonPath), hre);
     }
 
-    protected override get gatewayUrl(): string {
```
FabijanC, line 489: This shouldn't be removed. The adaptation only needs to take place if using DockerWrapper. Why would we adapt the URL if the user has Docker Desktop if they aren't even using the plugin in dockerized mode?

FabijanC (changes requested):

src/starknet-wrappers.ts
notV4l: my bad

src/utils.ts
notV4l: yeah :) but did u know https://www.bbc.com/future/article/20221011-how-space-weather-causes-computer-errors XD

FabijanC: that's it, I no longer need excuses when I can blame the universe xD

FabijanC (commented):
Looks better now, but we still have one unresolved conversation.

src/starknet-docker-proxy.ts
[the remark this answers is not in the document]
notV4l: to get a free port dynamically, i moved this call just before using this.port

FabijanC (approved):

PR #332, opened 2023-03-07
Declare return tx hash

### LINKED ISSUE #313: Class declaration should return transaction hash instead of class hash
Rationale:
- class hash can be retrieved in other ways (`contractFactory.getClassHash()`)
- there is no way of getting the transaction hash if it failed

This would be a breaking change.

TsBauer:
## Usage related changes

-  `declare()` now returns the tx instead of the class hash
- Now, the user needs to check the tx status for validating a successful declaration
- Closes #313 (breaking)

-   N/A

## Checklist:

-   [x] Formatted the code
-   [x] Linked issues which this PR resolves (#313)
    -   https://github.com/Shard-Labs/starknet-hardhat-example/pull/103
    -   [x] Restored `test.sh` to to use the original branch (after the example repo PR has been merged)

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -9,13 +9,13 @@ CONFIG_FILE_NAME="hardhat.config.ts"
 
 # setup example repo
 rm -rf starknet-hardhat-example
-EXAMPLE_REPO_BRANCH="plugin-declare-return-tx-hash"
-if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin-declare-return-tx-hash" ]]; then
+EXAMPLE_REPO_BRANCH="plugin"
+if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin" ]]; then
     echo "Invalid example repo branch: $EXAMPLE_REPO_BRANCH"
     exit 1
 fi
 
-git clone -b "$EXAMPLE_REPO_BRANCH" --single-branch git@github.com:brilliantblocks/starknet-hardhat-example.git
+git clone -b "$EXAMPLE_REPO_BRANCH" --single-branch git@github.com:Shard-Labs/starknet-hardhat-example.git
 cd starknet-hardhat-example
 git log -n 1
 npm ci
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -9,13 +9,13 @@ CONFIG_FILE_NAME="hardhat.config.ts"
 
 # setup example repo
 rm -rf starknet-hardhat-example
-EXAMPLE_REPO_BRANCH="plugin"
-if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin" ]]; then
+EXAMPLE_REPO_BRANCH="plugin-declare-return-tx-hash"
+if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin-declare-return-tx-hash" ]]; then
     echo "Invalid example repo branch: $EXAMPLE_REPO_BRANCH"
     exit 1
 fi
 
-git clone -b "$EXAMPLE_REPO_BRANCH" --single-branch git@github.com:Shard-Labs/starknet-hardhat-example.git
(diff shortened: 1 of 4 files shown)


src/types/index.ts
```diff
@@ -382,19 +382,9 @@ export class StarknetContractFactory {
...
 
         const executedOutput = executed.stdout.toString();
-        const classHash = extractClassHash(executedOutput);
         const txHash = extractTxHash(executedOutput);
 
-        return new Promise((resolve, reject) => {
-            iterativelyCheckStatus(
-                txHash,
-                this.hre.starknetWrapper,
-                () => resolve(classHash),
```
FabijanC, line 392: My idea was to keep on iteratively checking status, but eventually resolving not with `classHash` but with `txHash`, since the class hash can be obtained in different ways (`contractFactory.getClassHash()` as mentioned in #313).

Your change would make declare txs the only ones which don't wait for the transaction to become accepted. Ideally all transactions would behave the same (or is there an advantage in treating declaration differently?)

www/docs/intro.md
```diff
@@ -307,8 +308,13 @@ it("should estimate fee", async function () {
 it("should forward to the implementation contract", async function () {
     const implementationFactory = await starknet.getContractFactory("contract");
     const account = ...;
-    const implementationClassHash = await account.declare(implementationFactory);
+    const txHash = await account.declare(implementationFactory);
 
+    // ensure that tx has been successful
+    const txReceipt = await starknet.getTransactionReceipt(txHash);
+    expect(txReceipt.status).to.equal("ACCEPTED_ON_L2" || "ACCEPTED_ON_L1");
```
FabijanC, line 315: How should this work? Shouldn't some waiting have to be done? This is related to my other comment about potentially reverting to awaiting inside of `declare`.

FabijanC (changes requested):
The issue-linking mentioned in the checklist refers to actually making your PR automatically close the issue once it's merged to master. I achieved this by editing your PR description (added the keyword `Closes`).

Also, since the last time you contributed, we have enabled CircleCI workflow for outside contributors. I see a test of yours is failing (you can check it by clicking on :x:  here on the PR page). Probably you didn't await a declaration.

If some tests pass without awaiting their tx to become accepted, it's probably because they are run on Devnet, which by default accepts them upon receival. On real networks this acceptance is not instant.

src/types/index.ts
TsBauer: Fair point. Consistency is king. This will be changed👍

FabijanC:
@TsBauer @Dl-Vv I see that you applied some changes. Is your PR then ready for another round of review?

I would appreciate if you could paste relevant commit links as replies to my existing comments (can simply be copied from the PR discussion page, e.g. https://github.com/Shard-Labs/starknet-hardhat-plugin/pull/332/commits/b4f961d083b7b87abe130f6866df2f113b43fb3d - GitHub styles them like this automatically). I would also appreciate if you could then formally request my review here on GitHub when you are ready.

Thanks

src/types/index.ts
TsBauer: Re-introduced iterative checking in  3662a4435bca6a25e6c0193d87dd5d5a45474c27.

TsBauer:
@FabijanC Got it. We'll make it a habit👍

scripts/test.sh
```diff
@@ -9,13 +9,13 @@ CONFIG_FILE_NAME="hardhat.config.ts"
 
 # setup example repo
 rm -rf starknet-hardhat-example
-EXAMPLE_REPO_BRANCH="plugin"
-if [[ "$CIRCLE_BRANCH" == "master" ]] && [[ "$EXAMPLE_REPO_BRANCH" != "plugin" ]]; then
+EXAMPLE_REPO_BRANCH="plugin-declare-return-tx-hash"
```
FabijanC, line 12: Thanks for appropriately modifying test.sh. You can revert it now that https://github.com/Shard-Labs/starknet-hardhat-example/pull/103 has been merged.

FabijanC (approved):
Looks good, thanks for the contribution! Just revert the repo and branch used in `test.sh`.

scripts/test.sh
TsBauer: Done: 0d5a6c8d8e6f8bc08accf8911281fb5baa3f4a60

TsBauer:
It's a pleasure to contribute🙏

PR #342, opened 2023-04-20
Create LICENSE

ivpavici:


FabijanC (commented):
Should we change the license in package.json?

FabijanC (approved):

PR #349, opened 2023-05-02
utils.ts: fix finding constructors when a comment exists

cfal:
constructors aren't found correctly if a comment exists before it.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -397,7 +397,6 @@ export function isEntryAContructor(
             line = line.trim();
             if (line.startsWith("//")) {
                 // Ignore single-line comment.
-                continue;
             } else if (line.includes("#[constructor]")) {
                 // Check if next line is contains entry type name
                 const nextLine = lines[index + 1];


FabijanC (approved):

src/utils.ts
```diff
@@ -397,6 +397,7 @@ export function isEntryAContructor(
             line = line.trim();
             if (line.startsWith("//")) {
                 // Ignore single-line comment.
+                index++;
                 continue;
```
FabijanC, line 401: Since we haven't merged yet, I'll take the opportunity to suggest that instead of doing
```typescript
index++;
continue;
```
we simply do nothing. I.e. the `if` body should just be the comment `// Ignore single-line comment`, and the `index` increment will happen regardless at the end of the loop body.

FabijanC (changes requested):

FabijanC:
To better test this, #350 is created

FabijanC (approved):

PR #368, opened 2023-05-16
Update adaptInputUtil function return type

### LINKED ISSUE #356: Contract function calls shouldn't force returning an object
Currently in adapt.ts, if the function returns a raw primitive, we still force it to be wrapped into an object with property `"response"`:

https://github.com/0xSpaceShard/starknet-hardhat-plugin/blob/02c76812c6e90bd337f562f363ef9877099ec12f/src/adapt.ts#L353

We could instead return the value itself, which would require changing the return type from `StringMap` to `any` or `unknown` (or something else?)

Nathan-SL:
## Usage related changes

- Closes #356
### Breaking change
-   `adaptInputUtil` return type changed `StringMap` ->  `unknown`

-   NA

## Checklist:

-   [x] Formatted the code
-   [ ] Created a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/0xSpaceShard/starknet-hardhat-example):https://github.com/0xSpaceShard/starknet-hardhat-example/pull/111

### DIFF
--- a/src/adapt.ts
+++ b/src/adapt.ts
@@ -98,10 +98,6 @@ function convertOutputToU256(lo: bigint, hi: bigint): bigint {
     return (BigInt(hi) << BigInt(128)) | BigInt(lo);
 }
 
-function outputNameOrDefault(name?: string): string {
-    return name || "response";
-}
-
 // Can't use String.split since ':' also can be inside type
 // Ex: x : (y : felt, z: SomeStruct)
 function parseNamedTuple(namedTuple: string): starknet.Argument {
@@ -438,7 +434,7 @@ export function adaptOutputUtil(
     rawResult: string,
     outputSpecs: starknet.Argument[],
     abi: starknet.Abi
-): StringMap {
+): unknown {
     const splitStr = rawResult.split(" ");
     const result: bigint[] = [];
     for (const num of splitStr) {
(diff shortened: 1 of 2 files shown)


src/adapt.ts
```diff
@@ -452,15 +448,15 @@ export function adaptOutputUtil(
     for (const outputSpec of outputSpecs) {
         const currentValue = result[resultIndex];
         if (COMMON_NUMERIC_TYPES.includes(outputSpec.type)) {
-            adapted[outputNameOrDefault(outputSpec.name)] = currentValue;
+            adapted[outputSpec.name] = currentValue;
```
FabijanC, line 451: Don't we risk `outputSpec.name` to be undefined? If it's undefined, is it necessarily the only property? We need to have a cairo function in our tests which has an unnamed return parameter.

src/adapt.ts
```diff
@@ -438,7 +434,7 @@ export function adaptOutputUtil(
     rawResult: string,
     outputSpecs: starknet.Argument[],
     abi: starknet.Abi
-): StringMap {
+): unknown {
```
FabijanC, line 437: This seems like a very disruptive change. As much as I don't like `any`, but it would be less disruptive.

Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/368/commits/ad4f8352e5a1069ca34c86d09f56ade94f27448c

src/adapt.ts
Nathan-SL: So, from my investigation,
* In case of cairo 0
   - `outputSpec.name` wouldn't be `undefined` because any external function's return type must be named.

* In case of cairo 1
   - `outputSpec.name` could be `undefined` if it's the only property, `contract1.cairo` on our example repo has `get_balance` func with an unnamed return type.

FabijanC (approved):
Let's just restore the example repo branch

PR #370, opened 2023-05-17
Refactor constructor checking

### LINKED ISSUE #367: Refactor constructor checking
https://github.com/0xSpaceShard/starknet-hardhat-plugin/blob/18363cc4579895de69f0d2d11329893e9680754a/src/types/index.ts#L379

Plan:
- if constructors array is empty, we return a dummy always-false predicate 
- otherwise (meaning if length == 1, since we earlier checked if > 1), we return a predicate comparing with this only constructor name/selector 

We can also consider dropping the predicate approach, it's not like we'll have multiple predicates.

remiroyc:
## Usage related changes

-   N/A

-  closes #367  Refactor of the resolveConstructorPredicate() function. If the CONSTRUCTOR property is not declared in .casm, or if the value is an empty array, it will no longer produce an error. Instead, the function will immediately return a boolean false.

## Checklist:

-   [X] Formatted the code
-   [X] No linter errors + tried to avoid introducing linter warnings
-   [X] Performed a self-review of the code
-   [X] Rebased to the last commit of the target branch (or merged it into my branch)
-   [ ] Created a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/0xSpaceShard/starknet-hardhat-example):
    -   < EXAMPLE_REPO_PR_URL > 

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -402,7 +402,7 @@ export class StarknetContractFactory {
 
         // Can be simplified once starkware fixes multiple constructor issue.
         // Precomputed selector can be used if only 'constructor' name allowed
-        const selector = constructors[0]?.selector;
+        const selector = constructors[0].selector;
         return (abiEntry: starknet.AbiEntry): boolean => {
             return hash.getSelectorFromName(abiEntry.name) === selector;
         };
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -389,24 +389,25 @@ export class StarknetContractFactory {
             throw new StarknetPluginError(msg);
         }
 
-        if (!casmJson?.entry_points_by_type?.CONSTRUCTOR) {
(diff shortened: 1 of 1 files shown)


src/types/index.ts
```diff
@@ -389,24 +389,25 @@ export class StarknetContractFactory {
...
-        if (casmJson.entry_points_by_type.CONSTRUCTOR.length > 1) {
+        if (constructors.length > 1) {
             const msg = "There can be at most 1 constructor.";
             throw new StarknetPluginError(msg);
         }
 
         // Can be simplified once starkware fixes multiple constructor issue.
         // Precomputed selector can be used if only 'constructor' name allowed
-        const selector = casmJson.entry_points_by_type.CONSTRUCTOR[0]?.selector;
+        const selector = constructors[0]?.selector;
```
FabijanC, line 405: You can remove the question mark here, right?

FabijanC (changes requested):

Issue #140, opened 2022-07-08
Create a starknet.js wrapper


ivpavici:
Similar to how there are venv wrapper and docker wrapper, there should be a starknet.js wrapper and we should be moving toward relying on starknet.js instead of Starknet CLI

ivpavici:
closed since the CLI is being removed in favor of starknet.js in https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/400

PR #400, opened 2023-07-11
Replace internal Starknet CLI usage with starknet.js

penovicp:
## Usage related changes

- For happiest path scenarios the hope is that users should not be affected
- Minor API changes for ancillary functionalities and functions, most notably a truncated response for the `getBlock` method
- Some internally re-thrown errors might be different due to differences in the underlying tool used

- Starknet CLI internal usage removal in favour of starknet.js (`~5.17.0`) utilization
- new `starknetJs` property available for the runtime environment
- Resolves: #395 
- Resolves: #140 

## Checklist:

-   [x] Formatted the code
-   [x] Created a PR to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/0xSpaceShard/starknet-hardhat-example):
    -   https://github.com/0xSpaceShard/starknet-hardhat-example/pull/123

### DIFF
--- a/src/account-utils.ts
+++ b/src/account-utils.ts
@@ -1,10 +1,8 @@
 import axios, { AxiosError } from "axios";
-import crypto from "crypto";
-import { ec } from "elliptic";
 import fs from "fs";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import path from "path";
-import { ec as ellipticCurve, hash, number } from "starknet";
+import { ec, hash, stark } from "starknet";
 
 import {
     ABI_SUFFIX,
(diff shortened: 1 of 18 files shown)


src/account-utils.ts
```diff
@@ -34,31 +33,31 @@ export type CallParameters = {
...
     publicKey: string;
     privateKey: string;
-    keyPair: ec.KeyPair;
 };
 
 /*
  * Helper cryptography functions for Key generation and message signing
  */
-
+// _TODO: safe to deprecate?
```
FabijanC, line 41: Yes, safe to deprecate. This is not a documented feature of the plugin.

FabijanC (commented):
Looks good so far, thank you for the effort. I left one comment and updated the development changes section of the PR description to reflect the starknet.js update, feel free to edit it.

src/starknet-wrappers.ts
```diff
@@ -173,51 +152,6 @@ export abstract class StarknetWrapper {
 
     public abstract compileSierraToCasm(options: SierraToCasmOptions): Promise<ProcessResult>;
```
FabijanC, line 153: Is compilation the only reason why we need the StarknetWrapper logic?

I'll just mention this here: https://github.com/0xSpaceShard/starknet-hardhat-plugin/issues/387
So that the issue has a link to this comment, we don't have to deal with the issue here.

src/index.ts
```diff
@@ -174,6 +172,8 @@ function setVenvWrapper(hre: HardhatRuntimeEnvironment, venvPath: string) {
 
 // add venv wrapper or docker wrapper of starknet
 extendEnvironment((hre) => {
+    hre.starknetJs = new StarknetLegacyWrapper(hre.config.starknet.networkConfig);
```
FabijanC, line 175: Is the new `starknetJs` property of `hre` meant to be used both internally and by users?

FabijanC (commented):
What about documentation updates? Are there any? Were you planning on adding them?

src/starknet-wrappers.ts
penovicp: `starknet.js` could also be used to replace the hashing utilities although I believe it would be less performant.

penovicp:
> What about documentation updates? Are there any? Were you planning on adding them?

I wasn't planning to since the outward facing API should basically be the same (there are some minor differences for `getBlock`). It might make sense to note the CLI deprecation somewhere in the dev section of the documentation.

FabijanC (approved):
Basically ready, just the example repo branch specification needs to be restored after https://github.com/0xSpaceShard/starknet-hardhat-example/pull/123 is merged into that repo's `plugin` branch.

PR #401, opened 2023-07-17
Validate config file on test

### LINKED ISSUE #391: Add hardhat config file checking
While integrating scarb, I used this in the run_test function of test.sh:
```bash
    npx hardhat --typecheck --config "$CONFIG_FILE_NAME" 1>/dev/null ||
        echo "Invalid config file!"
```

We need a better version of this:
- exit after reporting an invalid config file
- restore the state as it was before the test (or don't change it at all - I was changing it by replacing the config file because I couldn't find another way of validating it).

Nathan-SL:
## Usage related changes

-   Closes #391

-   Validates config file on test

## Checklist:

-   [x] Formatted the code

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -59,8 +59,8 @@ function run_test() {
     else
         echo "Invalid config"
         # Clean up the temporary config file
-        rm -f "$CONFIG_FILE_NAME"
-        exit 1
+        rm "$CONFIG_FILE_NAME"
+        return 0
     fi
 
     # check if test_case/check.ts exists
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -53,7 +53,15 @@ function run_test() {
 
     # replace the dummy config (CONFIG_FILE_NAME) with the one used by this test
     /bin/cp "$config_file_path" "$CONFIG_FILE_NAME"
-    # in the future - validate config file here
+    # Validate config file
+    if npx hardhat --typecheck --config "$CONFIG_FILE_NAME" 1>/dev/null; then
+        echo "Config file is valid"
+    else
+        echo "Invalid config"
+        # Clean up the temporary config file
+        rm -f "$CONFIG_FILE_NAME"
+        exit 1
+    fi
 
     # check if test_case/check.ts exists
     if [ -f "$test_case/check.ts" ]; then


scripts/test.sh
```diff
@@ -53,7 +53,15 @@ function run_test() {
...
     /bin/cp "$config_file_path" "$CONFIG_FILE_NAME"
-    # in the future - validate config file here
+    # Validate config file
+    if npx hardhat --typecheck --config "$CONFIG_FILE_NAME" 1>/dev/null; then
+        echo "Config file is valid"
+    else
+        echo "Invalid config"
+        # Clean up the temporary config file
+        rm -f "$CONFIG_FILE_NAME"
+        exit 1
```
FabijanC, line 63: exit or return? Won't exit cause the whole script to stop, but we want to run all the test?

scripts/test.sh
```diff
@@ -53,7 +53,15 @@ function run_test() {
...
     # replace the dummy config (CONFIG_FILE_NAME) with the one used by this test
     /bin/cp "$config_file_path" "$CONFIG_FILE_NAME"
-    # in the future - validate config file here
+    # Validate config file
+    if npx hardhat --typecheck --config "$CONFIG_FILE_NAME" 1>/dev/null; then
+        echo "Config file is valid"
+    else
+        echo "Invalid config"
+        # Clean up the temporary config file
+        rm -f "$CONFIG_FILE_NAME"
```
FabijanC, line 62: Why `-f`?

FabijanC (changes requested):
Are we sure that tests are not affecting each other? I.e. if one test's config file is not valid, will this have effect no the following test case?

scripts/test.sh
Nathan-SL: [c2077b6](https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/401/commits/c2077b6ebd6df6dd642cde161b03518fe5dd8246)

scripts/test.sh
Nathan-SL: [c2077b6](https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/401/commits/c2077b6ebd6df6dd642cde161b03518fe5dd8246)

FabijanC (approved):

PR #402, opened 2023-07-19
Compiler downloader

### LINKED ISSUE #398: Replace dockerized cairo compiler by relying on prebuilt GitHub binaries
Since we are removing the Starknet CLI dependency, we can also remove it from our docker image. This means the Cairo compiler would remain the only useful thing in the image. We can also remove that and completely stop depending on the docker image, effectively terminating the need for [that extra repo](https://github.com/0xSpaceShard/cairo-cli-docker).

The binaries available [here under Assets](https://github.com/starkware-libs/cairo/releases) could be downloaded depending on the version specified by the user in their hardhat config file. We would somewhere download the specified binary (the exact asset depending on the user's machine), unpack it, and use it. We could use hardhat's Solidity handling for reference (mainly to see where they download the binaries, if they do so).

Nathan-SL:
## Usage related changes

- Closes #398
- If the user specifies a custom cairo1BinDir, this is used. If the `compilerVersion` is specified in the hardhat config file, this is used. These two options cannot be specified at the same time (there is a runtime check). If the user doesn't specify any, the last compiler version compatible with the plugin is used.
- If the user specifies a compilerVersion and it is not present, we download it for the user and store it in `~/.starknet-hardhat-plugin`. If already present here (and executable), we use that.

- Removed condition (and indentation) in setup-compiler.sh
- Extract common testing functionality by introducing `assertCompilationArtifactsExist`
- In `starknet-wrapper.ts`, the compilation functions are moved to the abstract parent class (`StarknetWrapper`)
  - The usefulness of starknet wrapper keeps on decreasing.

## Checklist:

-   [x] Formatted the code
-   [x] Created a PR (Wont affect this PR) to the `plugin` branch of [`starknet-hardhat-example`](https://github.com/0xSpaceShard/starknet-hardhat-example):https://github.com/0xSpaceShard/starknet-hardhat-example/pull/120


src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
-    protected getDockerArgs(): BindsMap {
-        const binds: BindsMap = {};
-        for (const source of this.sources) {
-            binds[source] = source;
+    async initialize(rootPath: string, starknet: StarknetConfig) {
+        this.compilerDownloadPath = starknet?.cairo1BinDir || path.join(rootPath, "cairo-compiler");
+        this.compilerVersion = starknet?.compilerVersion || (await this.getCompilerVersion());
+    }
+
+    async handleCompilerDownload(): Promise<void> {
```
FabijanC, line 42: It is not clear from the method name what it is doing. It could be `ensureCompilerVersionPresent`.

src/task-actions.ts
```diff
@@ -164,7 +165,12 @@ export async function starknetCompileCairo1Action(
     args: TaskArguments,
     hre: HardhatRuntimeEnvironment
 ) {
-    const binDirPath = getCompilerBinDir(args, hre.config.starknet);
+    let binDirPath = getCompilerBinDir(args, hre.config.starknet);
+    if (!binDirPath) {
+        const downloader = new CairoCompilerDownloader(hre.config.paths.root, hre.config.starknet);
+        await downloader.handleCompilerDownload();
+        binDirPath = downloader.getBinDirPath();
```
FabijanC, line 172: Ideally this would be just one call. I.e. some get method should ensure the compiler version is downloaded and return the path after that.

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+        await this.isInitialized;
+        if (fs.existsSync(this.getBinDirPath())) {
```
FabijanC, line 44: Here's the check I think we should be doing:
- get the binary path
- check if the binary file exists
- check if the binary file is executable by running the `starknet-compile --version` command, but simply relying on the exit code of the process being 0 (no str matching)
  - doing `output.contains(str)` can be wrong (e.g. `1.1.1` is a subset of `1.1.10`)

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+        }
+    }
+
+    public async getCompilerVersion(): Promise<string> {
+        const config = await import("../config.json");
+        return config["CAIRO_COMPILER"];
+    }
+
+    public getBinDirPath(): string {
+        return path.join(this.compilerDownloadPath, "target/release");
```
FabijanC, line 127: What is the point in using these target and release subdirectories?

src/task-actions.ts
```diff
@@ -164,7 +165,12 @@ export async function starknetCompileCairo1Action(
     args: TaskArguments,
     hre: HardhatRuntimeEnvironment
 ) {
-    const binDirPath = getCompilerBinDir(args, hre.config.starknet);
+    let binDirPath = getCompilerBinDir(args, hre.config.starknet);
+    if (!binDirPath) {
+        const downloader = new CairoCompilerDownloader(hre.config.paths.root, hre.config.starknet);
```
FabijanC, line 170: I don't think we should be storing the compilers in the project root. How does hardhat handle this with solidity?

Benefits of not storing in the project:
- Can be used across projets if the users has multiple on their machine (no need to download per project)
- The user won't have to edit their .gitignore to include the binaries

FabijanC (changes requested):
Does it work for compiling files outside of the repo? Was this case supported with older compiler?

www/docs/intro.md
```diff
@@ -544,13 +559,14 @@ module.exports = {
...
 
-If you're using `dockerizedVersion`, it will also use the dockerized Cairo 1 compiler version (currently 1.1.0). To specify your custom Cairo 1 compiler, you need to provide the path to the directory with its binary executables (likely a subdirectory of the target directory of your compiler repo). This can be configured in `hardhat.config.ts` or in [the CLI](#starknet-compile).
+To [compile](#starknet-compile) Cairo 1 files the plugin will download a compiler it's adapted to. You can also specify your custom Cairo 1 compiler, you need to provide the path to the directory with its binary executables (likely a subdirectory of the target directory of your compiler repo) or simply provide the version of the compiler you need and continue with compilation. This can be configured in `hardhat.config.ts` or in [the CLI](#starknet-compile).
 
 ```typescript
 module.exports = {
     starknet: {
         // if you build with `cargo build --bin starknet-compile --bin starknet-sierra-compile --release`
-        cairo1BinDir: "path/to/compiler/target/release/"
+        cairo1BinDir: "path/to/compiler/target/release/", // Plugin will prioritize this
```
FabijanC, line 568: And another thing, only one of these should be allowed. So similar to how we make a check for `venv` vs `dockerizedVersion`, we should do the same for `cairo1BinDir` and `compilerVersion`

Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/ff3d7491f473912f0966975b3d4afc658f14017b

src/cairo1-compiler.ts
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/4c7b88ba9de7e70a61d20380d7676248161e95ee

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+    }
+
+    public isDirEmpty(dirPath: string): boolean {
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/0db99f9b9cdb8bf69b177c4cd59b5cf979e00a18

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+    async initialize(rootPath: string, starknet: StarknetConfig) {
+        this.compilerDownloadPath = starknet?.cairo1BinDir || path.join(rootPath, "cairo-compiler");
+        this.compilerVersion = starknet?.compilerVersion || (await this.getCompilerVersion());
+    }
+
+    async handleCompilerDownload(): Promise<void> {
+        await this.isInitialized;
+        if (fs.existsSync(this.getBinDirPath())) {
+            // Checks if installed binary version is same as version set on hardhat config file
+            const isSameVersion = exec(`${this.getBinDirPath()}/starknet-compile --version`)
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/e6e17445bd16069ed9bf846b69ed366f05a6a026

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+            // Remove zip file after successfully extracting it
+            fs.rmSync(zipFile);
+        } catch (error) {
+            const parent = error instanceof Error && error;
+            throw new StarknetPluginError("Error extracting tar file:", parent);
+        }
+    }
+
+    public async getCompilerVersion(): Promise<string> {
+        const config = await import("../config.json");
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/81bf0eeb0056c6186c53f318a9587a5e84721fa3

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+
+            // Remove zip file after successfully extracting it
+            fs.rmSync(zipFile);
+        } catch (error) {
+            const parent = error instanceof Error && error;
+            throw new StarknetPluginError("Error extracting tar file:", parent);
+        }
+    }
+
+    public async getCompilerVersion(): Promise<string> {
```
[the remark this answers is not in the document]
Nathan-SL: Now removed

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+            // Checks if installed binary version is same as version set on hardhat config file
+            const isSameVersion = exec(`${this.getBinDirPath()}/starknet-compile --version`)
+                .stderr.toString()
+                .includes(this.compilerVersion);
+            if (isSameVersion) return;
         }
 
-        return binds;
+        // Check machine type
+        const fileName = os.platform() === "linux" ? FileName.LINUX : FileName.MACOS;
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/2f2015ce4082630fd5dfe766e85934a02d4ddbca

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+            throw new StarknetPluginError("Error downloading file:", parent);
         }
+    }
+
+    async extractZipFile(fileName: FileName): Promise<void> {
+        try {
+            const zipFile = path.join(this.compilerDownloadPath, fileName);
+            const corelibPath = path.join(this.compilerDownloadPath, "target/corelib");
+            // Execute the tar command to extract the tar/gz file
+            exec(`tar -xvf ${zipFile} -C ${this.compilerDownloadPath} --strip-components=1`);
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/28813b28a71615387267f2716e6158a67a9e870d

src/cairo1-compiler.ts
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/5efc50a7c425e58b74d93a39d4609ca0e4ecdc76

src/cairo1-compiler.ts
Nathan-SL: So, I avoided string matching and now it checks if the version is strictly equal to what is specified by the user. We need this because there could be an older/different version of the compiler installed already but different one in the `hardhat.config.ts` file.

https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/5efc50a7c425e58b74d93a39d4609ca0e4ecdc76

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
-        const binds: BindsMap = {};
-        for (const source of this.sources) {
-            binds[source] = source;
+    async initialize(rootPath: string, starknet: StarknetConfig) {
+        this.compilerDownloadPath = starknet?.cairo1BinDir || path.join(rootPath, "cairo-compiler");
+        this.compilerVersion = starknet?.compilerVersion || (await this.getCompilerVersion());
+    }
+
+    async handleCompilerDownload(): Promise<void> {
+        await this.isInitialized;
```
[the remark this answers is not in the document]
Nathan-SL: This is now removed due to use of `import` at the top of the file. https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/81bf0eeb0056c6186c53f318a9587a5e84721fa3

src/cairo1-compiler.ts
```diff
@@ -14,48 +25,110 @@ export const exec = (args: string) => {
...
+        if (fs.existsSync(this.getBinDirPath())) {
+            // Checks if installed binary version is same as version set on hardhat config file
+            const isSameVersion = exec(`${this.getBinDirPath()}/starknet-compile --version`)
+                .stderr.toString()
```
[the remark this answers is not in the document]
Nathan-SL: https://github.com/0xSpaceShard/starknet-hardhat-plugin/pull/402/commits/e23dd4ee2bda20eff2db5890a7002a2063c50f30

src/cairo1-compiler.ts
```diff
@@ -9,53 +22,140 @@ export const exec = (args: string) => {
...
-                binds: this.getDockerArgs()
+    async extractZipFile(fileName: FileName): Promise<void> {
+        try {
+            const zipFile = path.join(this.compilerDownloadPath, fileName);
+            const corelibPath = path.join(this.compilerDownloadPath, "target/corelib");
+            // Extract the tar/gz file
+            await tar.extract({
+                file: zipFile,
+                C: this.compilerDownloadPath,
+                strip: 1
```
FabijanC, line 119: What do these two things mean?

FabijanC (changes requested):
Needs conflict resolution. There are still three unresolved conversations, I marked the others as resolved.

src/cairo1-compiler.ts
Nathan-SL: - `C` sets the directory where the extracted files will be placed
- `strip` removes a specified number of leading directory when extracting, in this case 1

FabijanC: Can you add a comment in the code explaining why 1

FabijanC (approved):
Approving to make GitHub stop indicating that changes are requested. Assigning another reviewer.

penovicp (approved):

